PR #1, opened 2019-09-30
static build: Add workaround for nh2/static-haskell-nix#53

nh2:
See https://github.com/nh2/static-haskell-nix/issues/53

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,3 +1,9 @@
 resolver: lts-14.7
 packages:
 - .
+extra-deps:
+# Explicitly state these packages to make the static build (see default.nix) succeed.
+# Keep in sync with the versions provided by the `resolver` used above.
+# TODO: Remove when https://github.com/nh2/static-haskell-nix/issues/53 is fixed.
+- text-1.2.3.1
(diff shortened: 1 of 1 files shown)


david-janssen:
Thank you very much for this fix. I tried it and compilation is currently running. I have the sneaking suspicion this might take ages, and I'll let you know if it works out or not when it finishes, but I already merged this pull request, because it certainly got past the issue that I was having.

nh2:
> I have the sneaking suspicion this might take ages

In case you haven't already, enabling the `static-haskell-nix` binary cache (https://github.com/nh2/static-haskell-nix/#binary-caches-for-faster-building-optional) will ensure you don't have to build big things like GHC from scratch, only those packages from your `lts-14.7` that the binary cache hasn't built yet.

david-janssen:
Ooooh, great tip. I missed that. Added cachix and restarted compilation. I think I was trying to build multiple versions of ghc :-)

david-janssen:
Yeah, that finished and took no time at all. Additionally, I can easily run the binary in my own NixOS setup, or in an Ubuntu virtualbox.

Issue #4, opened 2019-10-06
Hack for openSUSE


wismill:
Hi there, thanks for the great work!

I had trouble to make kmonad work on openSUSE Tumbleweed. It does work using `sudo`, but if I want to run kmonad as normal user on my system there is no such group `uinput`. There is a group `input` but simply adding it to my user does not solve the issue.

My current solution is:
- create a group `uinput` and add my user to it
- add my user to the group `input`
- create a file `/lib/udev/rules.d/99-kmonad.rules` with the following content:
      ```
      KERNEL=="uinput", MODE="0660", GROUP="uinput", OPTIONS+="static_node=uinput"
      ```
- reboot

I am no expert. I wonder if this is secure and if it could be simplified?

slotThe:
I don't see why this would not be secure, afaik creating a group is the standard way to give users access to kernel modules (If this is wrong I would highly appreciate someone correcting me here).

As for simplifying, you don't have to create a `uinput` group if you already have `input`. I'm not sure what `OPTIONS+="static_node=uinput"` does but I'm fairly certain `MODE="0600"` is unecessary.  My `.rules` file simply contains `KERNEL=="uinput", GROUP="input"`, which seems to work fine so far.  Note that this is all from the perspective of a Void Linux install, I sadly don't have a Tumbleweed box at hand right now to test these wild speculations of mine.

wismill:
I removed the `uinput` group I created and change the rules files accordingly, but it requires the `OPTIONS` field to work:
```
KERNEL=="uinput", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput"
```

david-janssen:
I'm happy you managed to get things working. This isn't really a KMonad issue in the sense that KMonad needs to be adjusted to make this work, but I think it might be a good idea to create some documentation on how to get things running for different distros.

Should I just create an additional markdown file in the `doc` folder with further instructions for distros? You guys have an idea/opinion?

wismill:
Totally agree. Keep it simple, add a file and a link to it in the readme. This doc could then be completed with users' feedbacks from different distros.

david-janssen:
Very good. Won't have time for a few days, but will have a go in a couple of days. Once I do I'll close this issue.

slotThe:
I'm not sure creating a separate file is really necessary. I vote for just adding a note to the README, telling users to create a udev rule with
```
[...]

PR #5, opened 2019-10-06
Add README section about creating uinput group

slotThe:
"Fixes" #4 (by adding additional documentation).  

I'm not sure if it's worth telling users to check out the `input` group first (none of the system I've tried had `uinput` access with it).  Also not sure if we should advise them adding a whole new group instead of just reusing `input`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,21 @@ On most linux's this can be achieved by:
 sudo usermod -aG uinput username
 ```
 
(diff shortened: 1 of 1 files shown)


david-janssen:
Thanks, I'll merge this right away in case people run into this issue. If it turns out a lot of different distros need different configurations to get uinput to work we can always split it into a separate file then.

PR #9, opened 2019-10-11
Throw a config error when trying to define a non-existent layer

### LINKED ISSUE #8: Layer-toggling to non-existent layer should crash
Currently, it is possible to create a layer toggle button that switches to a non-existent layer. It is even possible to use this button once. Trying to access a key in the non-existent layer crashes KMonad, trying to switch to that layer again also crashes KMonad.

KMonad should simply refuse to start-up if the configuration contains a layer-switching button that points to a non-existent layer.

slotThe:
This closes #8 by throwing an error if trying to specify a non-existent layer as either an alias or directly inside the layer configuration.

Caveats:
  1. I don't know the code base (nor lenses) particularly well, so this might not be the most straightforward way to do things.
  2. `\\` has a worst-case complexity of `O(n^2)`, which could make this part very slow for big configs (I don't know if this is actually a realistic case but I thought I'd mention it anyways).  A possible solution would be to try and implement a fast version of `\\` (It seems possible to write an `O(n log n)` version, given we have an `Ord` instance for `ButtonToken` now).  Alternatively it might make sense to rewrite everything as a `Set`.
3. Error messages are not particularly good.  Ideally there should be some context, giving the user a hint where the error approximately occurs. It might even make sense to just show the first error, not a list.

### DIFF
--- a/src/KMonad/Core/Config.hs
+++ b/src/KMonad/Core/Config.hs
@@ -53,8 +53,10 @@ import Control.Exception
 import Control.Lens
 import Control.Monad.Except
 import Control.Monad.ST
+import Data.Containers.ListUtils (nubOrd)
 import Data.Foldable (toList)
 import Data.Hashable
+import Data.List ((\\))
 import Data.Maybe
 import Data.STRef
 
@@ -89,6 +91,7 @@ data ConfigError
   | AnchorNotFound      KeyCode Name
   | AlignmentError      Coor    ButtonSymbol Name
   | AliasNotFound       Symbol  Name
+  | LayerNotFound       [Name]
   deriving Exception
 
 instance Show ConfigError where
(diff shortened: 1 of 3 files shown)


david-janssen:
Sorry I didn't get a chance to review this today, had a full 24-hours of power-outage. I'll try to get to this soon! Thank you very much for contributing.

david-janssen:
Had a chance to review your code this morning. It is perfectly fit for purpose. In regards to your caveats:

1. This looks like the kind of solution I would've come up with myself.
2. I have purposefully told myself not to worry about performance *at all* in the parsing stages of KMonad, since that code only runs once on startup. Premature optimization and such. If we ever get all the major features up and running and multi-platform support etc, then we can always go back and shave some milliseconds from the startup time.
3. Yeah, all of the error messages in KMonad are awful :-). It's on my internal list, but below stuff like multi-platform support and improving documentation.

Thanks so much for contributing! I was just leaving a trail of breadcrumbs for myself to come back to (have been working on Windows support these last few days, when we had power), and it was a surprise that someone came along and just fixed my issue, but I guess that is what open source collaboration is. This is a lot of fun!

PR #19, opened 2019-11-16
KMonad is now packaged for Void Linux

slotThe:
I've packaged KMonad (version 0.3.0) for [Void Linux](https://voidlinux.org/).  If you want to look at the template file you can find it [here](https://github.com/void-linux/void-packages/blob/master/srcpkgs/kmonad/template).  
This pr just edits the README (and slightly restructures the `Getting KMonad` section) to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,17 @@ for an overview of different button-types.
 
 ## Getting KMonad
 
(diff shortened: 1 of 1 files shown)


david-janssen:
Sorry for the slight delay, just came back from holidays. Thank you very much for your contribution!

PR #47, opened 2020-08-05
Add layer-add and layer-rem buttons

obar:
This was in the documentation for the original version but I couldn't find these buttons in the refactor branch.


david-janssen:
I had actually left these 2 functions out on purpose, because I remembered them making it very easy to write broken configurations (i.e. configurations that contain ways to get into layer-situations you can't get out of again). But now you've added them again, and you know what, I am all in favor of handing the users a chainsaw, because a chainsaw, when handled well is a very useful tool.

Thanks for the contribution!

PR #68, opened 2020-09-06
Add 104-key template

obar:


david-janssen:
I'm not going to merge this request, because I simply forgot to put all the buttons into the 100% map. Thank you for catching on to my mistake, fixed in the current version of master :-)

obar:
Fair enough! I thought it was intentional with the 100-key file name :)

david-janssen:
100 for 100% :-)

PR #85, opened 2020-09-16
Activate fallthrough in testing.kbd

MaxGyver83:
I tested `testing.kbd` and got totally trapped. I couldn't stop KMonad because neither my mouse nor the control keys worked anymore. I could not even start a new terminal and execute `pkill kmonad`. So I had to poweroff my laptop the hard way.

I'm activating `fallthrough` for saving the next one trying out `testing.kbd`.


david-janssen:
I'll merge this because it's harmless, but I would like to point out that `testing.kbd` is just the config that I use for my own personal testing, which is evidenced by the fact that it lives in `keymap/user`.

MaxGyver83:
Alright! (I thought it was meant that everybody starts with another user's config and adapts it to their own liking.)

PR #88, opened 2020-09-17
Added a keymap template for ISO keyboards

precondition:
An alias "iso" for `KeyIso` has also been added.

I wasn't quite sure how to call the new keymap template file. At first, I named it `us_iso_60.kbd` in the same fashion as the preexisting `us_ansi_60.kbd` but then I thought that "US" and "ISO" in the same sentence sounds wrong. They don't care about the International Organization for Standardization in the United States. Another contestant was `uk_iso_60.kbd` but the keymap does not match the UK layout as it features the backslash on the ISO key. So I ended up with just `iso_60.kbd`

I've compiled my changes into a new executable and tested my keymap containing `iso` and it works fine.

### DIFF
--- a/keymap/template/iso_60.kbd
+++ b/keymap/template/iso_60.kbd
@@ -1,0 +1,27 @@
+#| --------------------------------------------------------------------------
+
+                          KMonad: ISO 60% template
+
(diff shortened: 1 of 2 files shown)


david-janssen:
Perfect, looks great

Issue #69, opened 2020-09-06
Syntax highlighters for Emacs/Vim


david-janssen:
If anyone is up to writing a syntax highlighter for `kbd` configuration files, that would be amazing :-)

WisdomWolf:
VSCode as well!

MaxGyver83:
I do `:set filetype=lisp` in vim. That's not perfect but still much better than the default highlighting.

precondition:
> I do `:set filetype=lisp` in vim. That's not perfect but still much better than the default highlighting.

The clojure syntax highlighter is a little nicer for kbd files, especially in regards to `t` in the keymap and button-macros

MaxGyver83:
> The clojure syntax highlighter is a little nicer for kbd files, especially in regards to `t` in the keymap and button-macros

You are right. It looks better with `:set filetype=clojure`. The biggest remaining issue is that the `"` character within `deflayer` blocks destroys the syntax highlighting from there until the end of the file. Escaping it fixes the syntax highlighting but KMonad doesn't accept `\"`. So I just replaced all occurrences of `"` with `S-'`. Now it looks good enough for me.

david-janssen:
Yeah, having issues here and there with those characters as well, maybe an alias like `quot` would be a good idea as well.

MaxGyver83:
Yes, I also searched for an alias. That would be less ambiguous than `S-'`.

Hugo-Heagren:
I've written a [vim syntax file](https://git.hugoheagren.com/Hugo-Heagren/KMonad.vim). I'm pretty sure it covers everything. This is my first time writing a syntax file though, so I'm not sure if I've got all the colour groups quite right. Let me know if there's anything I should change/add. It also includes syntax folding and matchit-jumping (<kbd>%</kbd> in vim) between multiline comment start/end markers.

I'm not sure how best it should be integrated into the kmonad world over all. I'm trying to move away from github to my own gitea instance, but that might not be great for something public which should be stable and accessible all the time. Maybe this is a good use for the kmonad organisation? I'd be happy to transfer ownership somehow to them, or have it in a repo owned by David or something. Whatever works...

habibalamin:
There's also [mwgkgk/kmonad-vim](https://github.com/mwgkgk/kmonad-vim).

slotThe:
@Hugo-Heagren

[...]

PR #89, opened 2020-09-18
kbd-mode for Emacs

slotThe:
This is in response to #69, trying to get some proper syntax highlighting working for Emacs.  ~~It's a [WIP] because, while I've been using it for some time and the basic functionality works great, some things turned out harder than expected and my elisp knowledge coming to it's end (also there's no documentation yet).  It's quite full of hacks but I suppose that's the elisp way :>~~

~~The basic highlighting (comments, keywords, etc.) is all done, the difficult part is getting quotes to play nicely.  Because in kmonad we often explicitly write `"` in the config file (and expect that to *not* be the start of a string) there are many situation where we don't actually want the "I'm inside a string right now" syntax highlight.  However, at times we *do* want that (e.g. when defining `uinput-sink`), so disabling syntax highlighting for quotes entirely does not sounds very desirable either.~~

~~Currently, it sort of works for `device-file`, but I don't quite know how to expand that regexp to the other candidates yet.~~

EDIT: Most of things should be working now.

### DIFF
--- a/elisp/kbd-mode.el
+++ b/elisp/kbd-mode.el
@@ -55,6 +55,14 @@
   '((t :inherit font-lock-variable-name-face))
   "Face for a variables, i.e. layer names, macros in layers,...")
 
+(defvar kbd-mode-show-string
+  '("uinput-sink" "device-file" "cmd-button")
+  "Syntax highlight strings in the S-expressions defined by these keywords")
+
+(defface kbd-mode-string-face
+  '((t :inherit font-lock-string-face))
+  "Face for strings")
+
 (defcustom kbd-mode-show-macros t
   "Whether to syntax highlight macros inside layout definitions.
 Default: t"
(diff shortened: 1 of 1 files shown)


slotThe:
@david-janssen I'm pretty happy with this at this point.  All of the issues mentioned in the OP should be fixed and I don't see anything critical missing; what do you think?

david-janssen:
Hey there,

This is great, well done :-). I'm going to have to find a chance to test it, and then we should definitely publish this somewhere. Do you think it best to include it with KMonad itself? Would it be better to put it in a new repo separately?

slotThe:
On Tue, Sep 22 2020 02:31, David Janssen wrote:
> Do you think it best to include it with KMonad itself? Would it be
> better to put it in a new repo separately?

Ah good question, I didn't think about this very much; it seems to vary
quite a lot from project to project as well.  For example, Agda and
hindent both have Emacs modes as part of the main repository, while the
TOML project just has some links in its wiki for all the different
editor modes.

I guess having separate repositories would be better in terms of
separating issues and pull requests, so I would be slightly in favour of
that.

david-janssen:
Cool, then I think we should make a new repo. Would you like me to host it on my account and add you as a maintainer, or would you like to host it on your own account and we link to it in the README? Either way works for me.

slotThe:
On Wed, Sep 23 2020 01:47, David Janssen wrote:
> Cool, then I think we should make a new repo. Would you like me to
> host it on my account and add you as a maintainer, or would you like
> to host it on your own account and we link to it in the README? Either
> way works for me.

Alright, I'll make a new repo on my github account then if you don't
mind.  I'll close this for now and open a new pull-request with the
documentation change once everything's set up :)

david-janssen:
Very good, I'll close this.

PR #107, opened 2020-09-30
add tutorial documentation on defsrc

focusaurus:
Not sure if others will jive with my approach, but here's a tutorial.

### DIFF
--- a/doc/tutorial-defsrc.md
+++ b/doc/tutorial-defsrc.md
@@ -1,0 +1,121 @@
+# Tutorial: Understanding defsrc and deflayer
+
+In this tutorial we'll create tiny `defsrc` and `deflayer` configurations to get a really solid understanding of how these work.
+
+## Prerequisites: Working kmonad and defcfg
+
+Please ensure you have kmonad installed and ready to run and you have figured out a `(defcfg)` expression in your config that works properly with your keyboard. This tutorial is going to focus exclusively on the `(defsrc)` and `(deflayer)` sections of your `.kbd` configuration file.
+
+We'll show a complete `.kbd` file but you should understand that the example is from a linux system. Many settings within `defcfg` will be different depending on your specific keyboard and OS.
+
+## Remapping a single key
+
+- Create a file `learn.kbd` similar to the one below.
+
+```
+(defcfg
+  input  (device-file "/dev/input/by-path/platform-i8042-serio-0-event-kbd")
+  output (uinput-sink "kmonad")
+  fallthrough true
+  allow-cmd false
+)
+
+(defsrc 1) (deflayer learn k)
+```
+- Start kmonad with: `kmonad learn.kbd`
+- Type your keyboard's `1` key
+
+You should see a letter `k` echoed by your terminal if you are still in the terminal with kmonad running, which is fine for the purposes of this tutorial. You can also switch to a text editor and you should see the same behavior.
+
+So now we understand a key thing: **Your defsrc does not need to represent every key on your keyboard**. In fact, it can be as simple as a single key. If all you want is to make your CapsLock key work as control, you can just configure that. It'll work on pretty much every keyboard, and you're done.
+
+We also now know how to do basic key remapping. Put an input key symbol in your `defsrc` and at the same positition (order) in your `deflayer` put the corresponding output key symbol.
+
+We also should now understand that the first `deflayer` entry in the file is enabled when kmonad starts.
+
+## Understanding fallthrough
+
+With kmonad still running, type other keys on your keyboard like letters, other numbers, etc. You should see them echoed in the terimal as you would expect. This is what is meant by `fallthrough true` in your `defcfg`. It just means kmonad will pass through unmodified anything else it sees from the input keyboard.
+
+One more thing to try: Hold `shift` as a modifier and type `1`. Do you get `!`, `K`, or something else? You should see `K` which is just another consequence of `fallthrough true`. kmonad passes through the shift key presses and when you press `1` with `shift` held kmonad sends `K`.
+
+## Remapping several keys in "order"
+
+Exit kmonad with `control+c`, change your `learn.kbd` to look as below and try that out.
+
+```
+(defcfg
+  input  (device-file "/dev/input/by-path/platform-i8042-serio-0-event-kbd")
+  output (uinput-sink "kmonad")
+  fallthrough true
+  allow-cmd false
+)
+
+(defsrc 1 2 3 4 5 6) (deflayer learn k m o n a d)
+```
+
+Now type `1 2 3 4 5 6`. You should see `k m o n a d`. You can remap as many keys as you like.
+
+Now let's change the order to solidify our understanding.
+
+```
+(defcfg
+  input  (device-file "/dev/input/by-path/platform-i8042-serio-0-event-kbd")
+  output (uinput-sink "kmonad")
+  fallthrough true
+  allow-cmd false
+)
+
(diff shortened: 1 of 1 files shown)


slotThe:
On Tue, Sep 29 2020 19:20, Peter Lyons wrote:
> Not sure if others will jive with my approach, but here&#39;s a tutorial.

Changes to the tutorial should go in `keymap/tutorial.kbd` I think; it
wouldn't make sense to have two tutorials in the same repository.

david-janssen:
Hey there, thanks for the pull request. I'd be happy to include your contribution, but I have a few questions first. I was wondering if there was a reason you didn't make clarifications to the `keymap/tutorial.kbd`? It sort of seems like you might've overlooked that file :-).

However! There is very much something to be said for a mark-downed tutorial. Whenever I sit down to write documentation for KMonad it always ends up being a lot of text, and it is hard to find a specific issue in the `tutorial.kbd`. I was thinking of starting a wiki, Github comes with built-in support for that, I think.

This is just me brainstorming, do you guys have any ideas for what would be best? In this particular case you are actually in a better position than me, since I've never had to approach KMonad from the outside, so it is hard for me to see what is clear and what is confusing.

david-janssen:
Oh, and more thoughts:

So I do want to avoid duplicating tutorials, 1 in `kbd` and 1 in`md` would be a headache to maintain. We would consider turning `kbd` into a 'feature-demonstration', and then move most of the text into `md` and link to `kbd`? But are 2 files easier to read than 1 big document? I don't know... Need input

focusaurus:
So my main goal here was to author some content. I'm not that particular about where it lands ultimately. I'll offer my opinion but I'm 100% fine if you prefer embedding the tutorial(s) in a `.kbd` file. In my opinion, for a project like this, I think all documentation should live in the README.md in the repo root including overview, installation, usage, etc. I'm fine if the file is very long. Here's some of my reasoning:

- It's the file often landed on after a web search. If everything's there, more folks will find everything.
- It's easy to click a link fragment and send someone a URL to a particular heading. Of course you can link to any line in any file on github but it's not as easy and not as stable as a link to a named fragment like `#troubleshooting` even as the document evolves.

And some of my reasoning why I don't prefer the "embed the docs as comments in the config file" approach

- Less conventional at this point vs a github readme.
- Less discoverable due to the above point plus SEO
- The `.kbd` file isn't even in the `docs` folder right now. It's like the docs are being hidden.
- The `.kbd` file extension is not well-known for "here be docs", nor would I expect to find tutorials in a directory named `keymap`

And while I did discover and read that `tutorial.kbd` file, for me, it's a recipe for the following:

- spend a long time reading that whole file and the evolution of advanced buttons I may or may not need
- Look at the samples and think "Before I can use or try this tool, I need to know the weird sym name for every key on my keyboard in order and if I get it wrong I might not be able to type on my computer"
- See that there's no existing file that matches my keyboard
- Try it out having exhausted my time box for futzing with this project, get an error message, and give up.

(I came very close to ending up at the end of that path)

So I wanted something that has a small feedback loop: read a little, run 1 command, get 1 result, either proceed or troubleshoot based on the result. I think markdown lends itself to guiding a user through that better than a comprehensive reference of the button syntax and capabilities.

david-janssen:
That is an incredibly cogent and convincing set of points you make. I will let this percolate for a bit, and will definitely take all of this into consideration.

My fear is that, as the configuration language and multi-platform options grow, the README will grow dauntingly large. What, in your experience, gives you the best interface to such a document? Just a good TOC? Or is a wiki-format at some point better?

I think also when it comes to coding, my brain stops wanting to contain files longer than like... 400 lines. But I think I might simply be a bit particular in this respect. 

@thoelze1 @slotThe @MaxGyver83 @eighal: I would love to hear your perspectives on this. After basic code-quality, documentation quality and accessibility is very important to me, but I honestly don't know what's best and am looking for opinions.

eighal:
I'm not an experienced project manager, but I have opinions about everything, so:

1. I think it's important to write documentation as you/we go, rather than after the fact, because after the fact I can never remember what I found puzzling.  You're already doing this so great!

2. I love wikis for everything provided that there's a friendly community (doesn't matter how big), which in this case there seems to be.  I have a poster of Ward Cunningham above my bed.  (Not really.  But almost.)

MaxGyver83:
My opinion:

* I prefer formatted documentation, with a table of contents and links to every section. But for me both Markdown or a wiki are fine.
* I think it's okay to split all documentation in separate Markdown files/wiki pages as long as everything is linked.
* Nevertheless it's good having some working example files containing some basic comments for getting started quickly. If somebody wants to do more than swapping some buttons in the example, he can check the detailed documentation.

slotThe:
In my opinion, the best solution would be some sort of literate
configuration file (as is often achieved by using org-mode).  In absence
of that I would choose a config file with comments as documentation over
markdown any day of the week.  I'm impartial on the wiki vs. config file
part.

On Wed, Sep 30 2020 07:22, Peter Lyons wrote:
> In my opinion, for a project like this, I think all documentation
> should live in the README.md in the repo root including overview,
> installation, usage, etc. I'm fine if the file is very long.

I think this only works for small README files and quite quickly falls
apart for larger and more complicated projects.  As an example, take
xmobar's README.md file (
https://github.com/jaor/xmobar/blob/master/readme.md ).  It's definitely
comprehensive, but I doubt anyone would call it easy to navigate (I
don't actually think that anything else (besides switching from markdown
to org-mode, of course!) would be better in xmobar's case, but that's
just because there are *so* many modules).

> - It's the file often landed on after a web search. If everything's
> there, more folks will find everything.

This, to me, sounds more like the README should give a short-ish
introduction into the topic instead of a deep-dive.  Give people a small
taste and watch them fall over themselves to get more :)

> - It's easy to click a link fragment and send someone a URL to a
> particular heading. Of course you can link to any line in any file on
> github but it's not as easy and not as stable as a link to a named
> fragment like `#troubleshooting` even as the document evolves.

I think this is fair, though it would point more towards a wiki than a
single README file.  Also, while you often only have a browser to read
the README (as you've most likely just discovered the project), you will
probably read the tutorial in a text editor of your choice, which would
make things like grepping for things rather easy.

> - The `.kbd` file isn't even in the `docs` folder right now. It's like
> the docs are being hidden.
> - The `.kbd` file extension is not well-known for "here be docs", nor
> would I expect to find tutorials in a directory named `keymap`

This I agree with, my first instinct when looking for the tutorial also
was to just go into the docs directory.  Though I think this would be
fixed rather easily.

> And while I did discover and read that `tutorial.kbd` file, for me,
> it's a recipe for the following:
>
> - spend a long time reading that whole file and the evolution of
> advanced buttons I may or may not need
> - Look at the samples and think "Before I can use or try this tool, I
> need to know the weird sym name for every key on my keyboard in order
> and if I get it wrong I might not be able to type on my computer"
> - See that there's no existing file that matches my keyboard
> - Try it out having exhausted my time box for futzing with this
> project, get an error message, and give up.
>
> (I came very close to ending up at the end of that path)

That is indeed rather unfortunate and I would agree that some sort of
"quick start guide" might be helpful to get people up and running
quickly.  At the end of the day though, kmonad does have a lot of
features and "hiding" those features from new adopters might be equally
frustrating to some.

david-janssen:
Going to close this pull-request because we need to basically rethink the entire concept of how to do documentation well. This isn't just a little command-line app, it's a keyboard configuration suite with a variety of buttons and config options.

Maybe we should expand the wiki, maybe we should make a read-the-docs, but a `tutorial.kbd` file isn't going to cut it.

PR #118, opened 2020-10-04
Consume whitespace after parsing a word

slotThe:
When applying the 'word' parser, it's often nice to be able to just ignore any
whitespace that follows.  This allows for nicely aligning things, like

```
(defalias
  layer1 (layer-toggle n                    )
  layer2 (layer-switch longername           )
  layer3 (layer-switch longestnamedefinitely))
```

As well as inserting comments in between these definitions

```
(defalias
  l1d (layer-delay
       ;; Some comment or other
       500
       ;; Another one
       layer1))
```

As we're only ignoring whitespace, this *shouldn't* interfere with any existing
configurations (I've tested it with a few of the user configs here and this does
indeed seem to be the case).


david-janssen:
Yeah, this makes perfect sense. Good catch!

Issue #113, opened 2020-10-02
Mappings for the "iso" key not working


stevep99:
I have an ISO keyboard with the extra key between the Z and left-shift.  As I understand it, this key should be referred to as "iso" in the configuration file.

Attached is a simple config file which demonstrates the issue. It re-arranges the left-hand bottom-row keys, such that Z is moved to the ISO key and so on. However any mapping that involves the ISO key produces no output. I am using the _**kmonad-0.4.1-linux**_ binary.

Expected result:
iso -> Z
Z -> X
X -> C
C -> V 
V -> B
B -> iso

Actual result:
iso ->  (no output)
Z -> X
X -> C
C -> V 
V -> B
B ->  (no output)

[test_iso.kbd.log](https://github.com/david-janssen/kmonad/files/5318951/test_iso.kbd.log)

david-janssen:
Are you sure that key is called `iso`? I think the key between `shift` and `z` on iso keyboards is called `102d`. The easiest way to check would be to run `evtest` on your input keyboard and see what it tells you about that key-name. Lemme know how it goes.

david-janssen:
Closing an reopening was completely unintentional. Faulty touchpad :)

stevep99:
Yes, can confirm 102d does work!

I was going by the [iso template file](https://github.com/david-janssen/kmonad/blob/master/keymap/template/iso_60.kbd) which calls it "iso". Also the the [Keycode.hs](https://github.com/david-janssen/kmonad/blob/master/src/KMonad/Keyboard/Keycode.hs) file has a reference to "iso" too.

david-janssen:
You are correct! We should fix the template file. It was user-submitted and I could never test it properly since I don't have the hardware. When I fix the documentation I'll close the issue. Thanks for the headsup.

In regards to the `Keycode.hs` file: yes, there is a KeyIso, although I don't know what actually triggers it :-)

MaxGyver83:
I was also wondering which key might be the `iso` key because `102d` works for me. Maybe it depends on the hardware?

precondition:
Oh yeah, you're right, my bad. `KeyIso` is mapped to `<I178>` in Linux which produces absolutely nothing. That keycode has no HID equivalent.

PR #122, opened 2020-10-06
Fixed ISO key

precondition:
Fixes issue #113 

As I've written in the comments of that issue, `KeyIso` is mapped to `<I178>` in Linux which produces absolutely nothing. Not even `NoSymbol` is assigned to it. That keycode has no HID equivalent. I guess `fallthrough` fooled me into thinking it works when I submitted PR #88


david-janssen:
Hey there,

Sorry for the late reply, things have been hectic. I am happy to pull the personal keymap change you made, but I don't think it is a good idea to make `iso` an alias for `102d`. If `KeyIso` isn't mapped to anything in X11, that's fine, but the name exists in the linux-headers, so 'KEY_ISO' is a thing, and I'd rather avoid the confusion. This is all *unless* you have a keyboard with actual ink on the key saying 'iso', in which case I'm happy to come up with some solution that takes this into account

If you could alter your pull-request, then I'd be happy to merge.

Cheerio,

D

precondition:
I doubt any keyboard has a key with "iso" written on it but that's generally how it is referred to in discussions.

What about `lsgt` as used in `xkbd` files or `nubs` as used in QMK? I'm not really a fan of the `102d` notation.

slotThe:
I have to say I quite like `102nd`, but the truth is that most 100% ISO
keyboards today do not have 102 keys anymore (having 105 instead) and so
the name may be nothing more than historical baggage at this point.

On Mon, Oct 12 2020 07:16, precondition wrote:
> What about `lsgt` as used in `xkbd` files or `nubs` as used in QMK?

I think both of those would work very nicely as aliases.

precondition:
> On Mon, Oct 12 2020 07:16, precondition wrote: What about `lsgt` as used in `xkbd` files or `nubs` as used in QMK?
> I think both of those would work very nicely as aliases.

Now that I think about it, it is true that nothing stops me from adding both

david-janssen:
Yeah, I'm absolutely fine with both `lsgt` and `nubs`, it's just that `iso` creates a semantic conflict with the Linux-headers we are mirroring that I would really like to avoid.

david-janssen:
Perfect! Thanks :)

Issue #90, opened 2020-09-22
Feature request: Universal config


MaxGyver83:
Meanwhile I've got a lot of KMonad config files. And many of them are almost identical.

So two things could help:

1. In Linux, it would be great if I could pass the `device-file` as command line parameter. Then I wouldn't need one config per keyboard (laptop/internal and USB/external).
2. It would be easy to have a shared config for Windows, Linux and MacOS: On Windows, `device-file` and `iokit-name` could be ignored, and `input (low-level-hook)` and `output (send-event-sink)` could be set automatically. On MacOS, `device-file` could be ignored and `iokit-name` could be considered instead.

thoelze1:
I like this idea but I should add that, unfortunately, the mac interface provides some OS-specific keycodes that (if present) would add complexity to a cross-compatible config

MaxGyver83:
That's my workaround for my Linux configs (it's a fish function it should be easy to adapt it for bash):

`~/.config/fish/functions/k.fish`:

```fish
function k --description 'Start KMonad for all available keyboards'
    set config ~/.config/kmonad/vou4.kbd
    pkill kmonad

    echo "Activate KMonad for internal keyboard"
    kmonad $config & ; disown

    set keyb /dev/input/by-id/usb-Telink_Wireless_Receiver-if01-event-kbd
    if test -e $keyb
        echo "Activate KMonad for Jelly Comb keyboard"
        set dest "/tmp/kmonad-vou-jelly.kbd"
        cp $config $dest
        sed -i 's:/dev/input/by-path/platform-i8042-serio-0-event-kbd:'$keyb':' $dest
        kmonad $dest & ; disown
    end

    set keyb /dev/input/by-id/usb-SONiX_USB_DEVICE-event-kbd
    if test -e $keyb
        echo "Activate KMonad for Royal Kludge keyboard"
        set dest "/tmp/kmonad-vou-royal.kbd"
        cp $config $dest
        sed -i 's:/dev/input/by-path/platform-i8042-serio-0-event-kbd:'$keyb':' $dest
        kmonad $dest & ; disown
    end
end
```

david-janssen:
Absolutely 100% agreed, and I opened a 'self-issue' to implement something like this at #62, although perhaps I should have been more verbose.

The solution I had in mind was the following:
1. All command-line options should have a corresponding entry in `defcfg`, so you can specify a default.
[...]

PR #137, opened 2020-11-10
[WIP] Create CLI for all Options in DefCfg

slotThe:
This is in partial response to the discussion in #90.  It basically (so far) implements the following:

> Any all defcfg entries should have command-line ways of specifying them, and these should always override defcfg if provided on the command line.

I marked this as a WIP because there are still some things I'm not sure about.

  1. What is the purpose of `SInitStr`?  Unless I'm going crazy, this seems to be just ignored in `KMonad.Args.Joiner`!  There are still TODOs in the source file for it for this reason.
  2. I summoned an (ugly) `Eq` instance for `DefSetting`, comparing only the outermost constructor.  This makes sense right now because we only allow a single e.g. output token to be specified in the config file.  Should this ever change then this instance would need to change with it (or, realistically, be deleted entirely).
  3. `joinCLI` is not the prettiest function, but so far I can't think of a better way to incorporate command line options into the config options while also maintaining certain invariants (like the number of config options specified, in case we have to bail out with a `DuplicateSetting` exception).

One could argue that 2. and 3. are connected.  If we allowed several options (and perhaps even `defcfg` blocks) to be specified, then a simpler implementation could be found.  However, if we don't just want to pick the first specified block but one that "works", the question arises what "works" even means in this context (and what if we have two specifications that work?).  I went with the "easy" way for now, because I quite frankly don't know the answer to a lot of these ambiguity questions (though I would be prepared to rewrite this if you do).   

At the moment, command line options may be specified in the following way:

```shell
kmonad -o "uinput-sink 'NAME' 'sleep 0.2s; xset r rate 230 70'" -f -c /path/to/config
```
Note the missing parenthesis around the output token, as well as the string being inside single quotes; I thought that this would be a decent quality-of-life improvement over using normal sexp syntax in the CLI.

### DIFF
--- a/src/KMonad/Args/Cmd.hs
+++ b/src/KMonad/Args/Cmd.hs
@@ -16,7 +16,11 @@ module KMonad.Args.Cmd
   )
 where
 
-import KMonad.Prelude
+import KMonad.Prelude hiding (try)
+import KMonad.Args.Parser (itokens, keywordButtons, noKeywordButtons, otokens, symbol)
+import KMonad.Args.Types (DefSetting(..), choice, try)
+
+import qualified KMonad.Args.Types as M  -- [M]egaparsec functionality
 
 import Options.Applicative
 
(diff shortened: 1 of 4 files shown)


david-janssen:
This looks great!

I haven't tested it yet, but since it's going into develop, I'm just going to approve this for now. I will tinker with some other stuff when time/energy/health permits, but I'm dreaming of a big refactor of the CLI anyways, this is a great start in that direction.

src/KMonad/Args/Cmd.hs
```diff
@@ -79,3 +100,64 @@ levelP = option f
...
+  <> metavar "STRING"
+  <> help "TODO"
+  )
+
+-- | Key to use for compose-key sequences
+cmpSeqP :: Parser (Maybe DefSetting)
+cmpSeqP = optional $ SCmpSeq <$> option
+  (tokenParser keywordButtons <|> megaReadM (choice noKeywordButtons))
+  (  long "cmp-seq"
+  <> short 'c'
```
MaxGyver83, line 133: I just tested the develop branch and noticed that the short argument `-c` is defined twice (= `--allow-cmd` and `--cmp-seq`).

Same for `-i`. This is short for `--init` and `--input`.

slotThe: Hah, you are right of course.  That's what I get for only testing the long options; fixed in #164

MaxGyver83: That was quick! :smile:

PR #145, opened 2020-11-27
Add a NixOS module

WhatisRT:
This adds a NixOS module that installs kmonad and runs it automatically using a systemd service.


david-janssen:
Thanks! This is great.

PR #146, opened 2020-11-28
Add ability to execute shell command on release of cmd-button

### LINKED ISSUE #25: Execute command on layer change
Sometimes it might be desirable to execute a shell command whenever the layer changes (e.g. if you wanted to change the input language of your keyboard on layer X).  I don't think this is possible right now.

I would try hacking this in myself but I'm not sure how extensive the planned refactoring will be.  I'd rather wait until it's (mostly) finished before trying to add new features (though I just realized that there is another branch for that and you haven't been doing all the development offline :o).

slotThe:
Closes #25. 

With this change, as well as the help of the `around` function, one is
able to use a `cmd-button` for e.g. switching the keyboard layout (via
`setxkbmap` or a similar utility) after toggling a layer and then
switching back to the normal layout after the toggle has been released.

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -68,8 +68,11 @@
     If this is set to `false`, any action that runs a shell-command will simply
     log to `stdout` without ever running (log-level info). Don't ever enable
     this on a configuration that you do not trust, because:
+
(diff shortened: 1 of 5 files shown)


david-janssen:
This is incredibly cool, well done :-)

PR #149, opened 2020-12-06
User config neo keyboard layout

MaxGyver83:
- Fix order of greek letters on layers 5 and 6 in `neo.kbd`
- Add `neo-de.kbd` which is based on a German (QWERTZ) keyboard layout. This config makes use of a lot of aliases. It could help prevent some problems because keys available in the German layout can be sent "directly"/"natively" without using composed characters (like `ä` or `ß`).


slotThe:
Thanks!

PR #152, opened 2020-12-09
[WIP] Delay in tap-macros/compose-key sequences

### LINKED ISSUE #127: Compose key sequences are unreliable
The input of Umlauten is very unreliable when using the compose key option with the default sequences. I was able to trace the issue back to a too fast execution of the sequence. After building my own with a 5 ms pause in it, it worked reliable and as expected.

```
(defcfg
  ;; For Linux
  input  (device-file "/dev/input/by-path/platform-i8042-serio-0-event-kbd")
  output (uinput-sink "KMonad output" "")
  fallthrough true
)

(defalias ü #(S-RA-P0 " P5 u))
(defsrc          o u)
(deflayer main ö @ü)
```

In this case ö would trigger only every so often and ü would run perfectly.

As I've set the compose key in the settings (Gnome Tweaks), there's no need for me to execute `sleep 1 && setxkbmap ...`.  

My system is an Ubuntu 20.04 LTS with custom 5.7.10 mainline kernel. I'm running on Wayland.

slotThe:
Fixes #127 

It's a bug fix so it's against `master`

Some systems (such as GNOME 3, apparently) have troubles when compose-key sequences are entered too fast, resulting in mangled output.  The added delay is so small (5 ms) that it should not be percetible for anyone.

Because this is such a tiny (and non-disruptive) change, I'll merge it in a couple of days if no one protests

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -83,6 +83,13 @@
     as a workaround to also make this work on windows, see the section on
     Compose-key sequences below.
 
+  - `cmp-seq-delay': NUMBER (in milliseconds)
+
+    This sets a delay between each pressed key in a compose-key sequence.  Some
+    environments may have troubles recognizing the key sequence if it's pressed
+    too rapidly; if you experience any problems in this direction, you can try
+    setting this value to `5' or `10' and see if that helps.
+
   Secondly, let's go over how to specify the `input` and `output` fields of a
   `defcfg` block. This differs between OS'es, and so do the capabilities of
   these interfaces.
@@ -147,7 +154,8 @@
     ;; To understand the importance of the following line, see the section on
     ;; Compose-key sequences at the near-bottom of this file.
     "/run/current-system/sw/bin/sleep 1 && /run/current-system/sw/bin/setxkbmap -option compose:ralt")
-  cmp-seq ralt  ;; Set the compose key to `RightAlt'
+  cmp-seq ralt    ;; Set the compose key to `RightAlt'
+  cmp-seq-delay 5 ;; 5ms delay between each compose-key sequence press
 
   ;; For Windows
   ;; input  (low-level-hook)
@@ -478,8 +486,19 @@
   the program time to register all the key-presses. Therefore we also provide
   the 'pause' function, which simply pauses processing for a certain amount of
   milliseconds. Pauses can be created like this:
+
     (pause 20)
     P20
+
+  You an also pause between each key stroke by specifying the `:delay' keyword,
+  as well as a time in ms, at the end of a `tap-macro':
+
+    (tap-macro K M o n a d :delay 5)
+    #(K M o n a d :delay 5)
+
+  The above would be equivalent to e.g.
+
+    (tap-macro K P5 M P5 o P5 n P5 a P5 d)
 
   WARNING: DO NOT STORE YOUR PASSWORDS IN PLAIN TEXT OR IN YOUR KEYBOARD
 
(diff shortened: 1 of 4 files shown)


david-janssen:
I have no issues with the fix 'an sich', although I would like to point out that we are mixing our abstractions. I.e. we are now hardcoding some text into a bit that will be parsed. Ideally we would create a timer argument into keyboard macros, and then make that timer argument somethign that people could pass in.

*However*, since the abstractions are already a bit mixed, and since a proper rework of compose-key sequences and keycodes is something we should have a go at anyway, I think this approach is temporary and not too invasive, so it gets my stamp of (reluctant) approval.

slotThe:
On Wed, Dec 09 2020 23:33, David Janssen wrote:
> I have no issues with the fix 'an sich', although I would like to
> point out that we are mixing our abstractions. I.e. we are now
> hardcoding some text into a bit that will be parsed. Ideally we would
> create a timer argument into keyboard macros, and then make that timer
> argument somethign that people could pass in.

Oh!  I did not think about this; that sounds like a much better
solution.  I think I'll rather have a go at that, being able to give
pauses directly to keyboard macros sounds like a useful thing to have

david-janssen:
I commend your commitment to clarity :-)

slotThe:
My pleasure :)

I have poked around in the code a bit today and will write down some preliminary thoughts here, just so I don't forget.  I intend to just force-push the changes to this branch, so this pr can stay open. 

Adding delay support to `tap-macro` would be a bit of a breaking change as far as I can tell.  The problem is that the list of supplied `DefButton`s can of course be arbitrarily long.  We could inspect the last button in that list and see if it's a `KPause`, but then we'd break macros that really only want a delay after the last button has been pressed (and I bet someone, somewhere, is relying on that). 

We could introduce a new `tap-macro-pause` function, but I'm not sure that's worth it just for this functionality. 

Since `KComposeSeq` can't be specified directly (only through dead-keys and the respective unicode symbols), we might just inject pauses into it; users wouldn't be able to control this part anyways.  This way, we wouldn't get the whole "`tap-macro`s now have this cool functionality", but we also wouldn't get any breaking changes/new functions.

I'm not sure what the right approach is here, or if there's a solution to all this that I'm missing at the moment.

david-janssen:
What about a global configurable parameter, like `compose-key`, that defines how many ms to pause between macro-steps? That way we can set it to 5 for gnome issues, leave it at 0 for no effect, and people can still easily add manual pauses as well.

slotThe:
That's an interesting suggestion; we'd gain the ability to explicitly set it for compose-key sequences, but users couldn't set it only for specific `tap-macro`s (which may not be a big deal after all).

Another idea that occured to me just now: Since we're pretending to be a list anyways, how about using record-accessor-like syntax for something like 
``` clojure
(tap-macro a b c d
           :delay 5)
```

This would again have the problem that compose-key sequences are only defined via the respective keys, so we'd have to hard-code some sort of tiny delay.

david-janssen:
Yeah, I was already hoping to prove that `:keyword value` style syntax for a variety of uses (also customizing layer-definitions, for example).

slotThe:
@david-janssen this now is a first draft that implements the `:delay` keyword for `tap-macro`s, as well as a new global parameter `cmp-seq-delay` to specify this for compose-key sequences.  What do you think? 

As this qualifies as a new features I think it should now be made against `develop` instead of master; I'll change that as soon as the pr has been finalized

david-janssen:
This looks really cool: I will have a look at this tomorrow, life permitting, and let you know :-)

slotThe:
@david-janssen I've rebased this on top of current `master`.  Would you mind taking a look?

PR #154, opened 2020-12-13
My config, to be added to the user configs

Rougemoot:


Rougemoot:
I discovered kmonad a couple days ago and immediately switched from Karabiner.

Having found myself in need of some example configs while writing mine, I thought I'd contribute by making my config publicly available with some explanation in a readme.

I hope this is useful to somebody, as said in the readme it has a bit of everything (I was surprised to see no home row modifiers examples in the available user configs even though the tutorial mentioned them).

Cheers!

slotThe:
This looks nice, thanks!

PR #162, opened 2020-12-29
change to multi-keyboard systemd units in nixos

stites:
A (breaking) change to spawn one kmonad instance for each config file a user has written.

this also ensures that the service is human-readable (ie: `kmonad-my.kbd`) -- I _think_ nixos won't let you write systemd units with identical names, so the case of two `my.kbd` configs at different paths is implicitly covered (and will error).

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -44,10 +44,19 @@ with lib;
         KERNEL=="uinput", MODE="0660", GROUP="uinput", OPTIONS+="static_node=uinput"
       '';
 
-    systemd.services = with lib; with builtins;
+    systemd = with lib; with builtins;
       let
+        make-group = length cfg.configfiles > 1;
+        wantedBy = [ "graphical.target" ];
+        mk-kmonad-target = services: {
+          description = "KMonad target";
+          requires = map (service: service.name + ".service") services;
+          inherit wantedBy;
+        };
         mk-kmonad-service = kbd-path:
-          let conf-name = lists.last (strings.splitString "/" (toString kbd-path));
+          let
+            conf-file = lists.last (strings.splitString "/" (toString kbd-path));
+            conf-name = lists.head (strings.splitString "." conf-file);
           in {
           name = "kmonad-" +conf-name;
           value = {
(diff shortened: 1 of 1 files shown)


bard:
Maybe relevant:

- http://0pointer.de/blog/projects/instances.html
- https://www.stevenrombauts.be/2019/01/run-multiple-instances-of-the-same-systemd-unit/

stites:
Thanks! I think templates are good for dynamically making units on-the-fly but in nix-based systemd units I've always seen them programmatically generated. It's very possible that I just don't know enough nix to do this. Maybe someone has some pointers for this? Otherwise I will keep this as-is 

The other thing suggested in these articles is adding a target, which I'll definitely add! It necessitates the question of "what happens in the case of one config file?" I'll try to do the intuitive thing (>1 unit will be added to a group, 1 unit will remain as-is) but the code might get a bit messy.

stites:
Another thing I would like to fix, but I'm not sure how: the systemd units will fail if, for instance, you rebuild nix without the device plugged in.

Maybe just an if-then statement before running kmonad?

slotThe:
This looks great, thanks!  I suppose the fact that this is a breaking change is not terribly bad, as the current nix config hasn't made it into a release yet.  

I personally don't use nix (so I can neither properly review this, nor advise you about the finer details here), so I'll wait for someone who does to greenlight this

david-janssen:
Was looking through this and it does look great indeed! However I think neither @slotThe or I is capable enough with nix to really judge (I use nix, but it's a real bodge-job). Would you mind explaining to me exactly what and how this will break things?

Maybe it's a good idea to merge it into develop instead, and when we've completed some more fixes do a new release?

nix/nixos-module.nix
```diff
@@ -14,12 +14,12 @@ with lib;
...
     };
 
-    configfile = mkOption {
-      type = types.path;
-      default = "";
-      example = "my-config.kbd";
+    configfiles = mkOption {
+      type = types.listOf types.path;
+      default = [];
+      example = "[ my-config.kbd ]";
```
stites, line 20: Aside from the pluralization of the option, this also takes a list of config paths. This was the breaking change i was referring to since someone might just pull in this repo from master and load this into their config via overlay.

stites:
I ran into another problem with the systemd setup and wrote a fix a few weeks ago. Namely, kmonad will throw a non-zero exit if the keyboard isn't plugged in when the service starts, so systemd looks broken half the time for me. Adding an "optional config" option is my current hack and it just appends `|| true` to the kmonad script.

I constantly forget nix functions as well and a lot of my nix knowledge just comes from reading nixpkgs source so I added some comments -- hopefully this clears up some of the magic.

Happy to move this to the develop branch if you'd like. Just let me know what you'd like to do!

david-janssen:
Yah, if you could put this on develop I'll merge it in right away. :-)

stites:
moved to #190

PR #171, opened 2021-01-13
Add systemd user unit

NullSense:
I noticed there's no systemd unit file in the repo so I thought I'd contribute with mine, I didn't know where to put it, but this is the first working version. Let me know if it interests you.

### DIFF
--- a/startup/kmonad.service
+++ b/startup/kmonad.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=kmonad keyboard config
+
+[Service]
+Restart=always
+RestartSec=3
+ExecStart=/usr/bin/kmonad %E/homemod.kbd
+Nice=-20
+
+[Install]
+WantedBy=default.target


slotThe:
On Wed, Jan 13 2021 02:37, NullSense wrote:

> I noticed there's no systemd unit file in the repo so I thought
> I'd contribute with mine, I didn't know where to put it, but
> this is the first working version. Let me know if it interests you.

This looks nice, thanks!  I can't properly review (nor test) this, as I
don't use systemd and hence don't know much about *how* one should write
these services; so I'll wait for others to chime in.

bard:
@NullSense Nice!

Instead of `%h/homemod.kbd`, you might want to have it read configuration from `%E/kmonad/config.kbd` (which will resolve to `~/.config/kmonad/config.kbd` for user units and `/etc/kmonad/config.kbd` for system units) so that people will be able to place the configuration file in the conventional places, without having to edit the unit file.

NullSense:
> @NullSense Nice!
> 
> Instead of `%h/homemod.kbd`, you might want to have it read configuration from `%E/kmonad/config.kbd` (which will resolve to `~/.config/kmonad/config.kbd` for user units and `/etc/kmonad/config.kbd` for system units) so that people will be able to place the configuration file in the conventional places, without having to edit the unit file.

Done!

bard:
I'm curious why you picked a `Nice=19`, i.e. lowest priority? I don't have a `Nice` setting in my unit (which is otherwise identical to yours) but if I had to pick one, I think I would have gone with the highest priority, being user input.

Other than that, as I said an almost identical one has been working well for me for a while, in case that gives you a data point @slotThe .

NullSense:
> I'm curious why you picked a `Nice=19`, i.e. lowest priority? I don't have a `Nice` setting in my unit (which is otherwise identical to yours) but if I had to pick one, I think I would have gone with the highest priority, being user input.
> 
> Other than that, as I said an almost identical one has been working well for me for a while, in case that gives you a data point @slotThe .

Actually negative numbers indicate higher prio, from `man nice`:

```
Run COMMAND with an adjusted niceness, which affects process scheduling.  With no COMMAND, print the current niceness.  Nice‐
       ness values range from -20 (most favorable to the process) to 19 (least favorable to the process).
```

(however -20 is not available with systemctl units for some reason)

bard:
> > I'm curious why you picked a `Nice=19`, i.e. lowest priority? I don't have a `Nice` setting in my unit (which is otherwise identical to yours) but if I had to pick one, I think I would have gone with the highest priority, being user input.
> > Other than that, as I said an almost identical one has been working well for me for a while, in case that gives you a data point @slotThe .
> 
> Actually negative numbers indicate higher prio, from `man nice`:

That's what I'm saying. :) I'd have expected to see a negative number as `Nice` value.

NullSense:
> > > I'm curious why you picked a `Nice=19`, i.e. lowest priority? I don't have a `Nice` setting in my unit (which is otherwise identical to yours) but if I had to pick one, I think I would have gone with the highest priority, being user input.
> > > Other than that, as I said an almost identical one has been working well for me for a while, in case that gives you a data point @slotThe .
> > 
> > 
> > Actually negative numbers indicate higher prio, from `man nice`:
> 
> That's what I'm saying. :) I'd have expected to see a negative number as `Nice` value.

Oh I misread my own code, nice catch :sweat_smile:. Changed it now,

slotThe:
@bard thanks!

@NullSense This seems simple enough to merge without a fire starting somewhere :)  What do you think of creating a `scripts` directory and putting the service in there?  Having a place for these kind of things would be a plan (maybe I'll even get around to adding a runit script!)

NullSense:
Maybe `startup` if you're also planning to add runit scripts? Unless there's more than daemon management that is going to be added.

slotThe:
On Tue, Jan 19 2021 15:38, NullSense wrote:

> Maybe `startup` if you're also planning to add runit scripts? Unless
> there's more than daemon management that is going to be added.

Sounds good to me; fits what's in it better at the moment.  Should we
really have the need to add more non-startup scripts at some point, we
can just rename the directory then and there

NullSense:
> On Tue, Jan 19 2021 15:38, NullSense wrote: Maybe `startup` if you're also planning to add runit scripts? Unless there's more than daemon management that is going to be added.
> Sounds good to me; fits what's in it better at the moment. Should we really have the need to add more non-startup scripts at some point, we can just rename the directory then and there

done @slotThe

slotThe:
Perfect, thanks!

PR #181, opened 2021-02-01
Use conventional XDG config directory

bard:
Minor change to the recent systemd unit addition: use `~/.config/kmonad/`  (or `/etc/kmonad/` if called as system service) as configuration directory instead of `~/.config/` (or `/etc/` if called as system service).


slotThe (approved):

slotThe:
Respecting XDG conventions sounds like a very sensible thing to do; will merge this in a few days if there are no objections.

PR #182, opened 2021-02-01
Add runit service

slotThe:
Now that we have a systemd unit, I can't let runit fall behind!

Also mention the startup scripts in the `README.md`, because it feels like something that we should do

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ Want to add your own keyboard configuration to `./keymap`? Fork KMonad, create a
 ### Editor Support for the Configuration Language
 - [Emacs](https://github.com/slotThe/kbd-mode)
 
+### Startup
+There are startup scripts available for different init systems in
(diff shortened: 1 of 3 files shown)


slotThe:
It's a tiny shell script (can't really butcher something so short) and I think the doc improvement is a net-positive, so I will merge this in a few days if there are now objections

david-janssen:
This is great. Could you add an entry to the changelog? Then we'll merge this in right away.

slotThe:
On Sun, Feb 07 2021 23:56, David Janssen wrote:
> This is great. Could you add an entry to the changelog? Then we'll
> merge this in right away.

Ah good point, it's done.  I also added a note for the systemd unit, as
that was forgotten in the respective pr as well.

david-janssen:
Perfect!

PR #190, opened 2021-02-22
allow for multi-keyboard services in nixos

stites:
Followup from https://github.com/david-janssen/kmonad/pull/162


slotThe:
I'll just merge this since it's against `develop`; thanks!

PR #204, opened 2021-03-27
installation: add link to Arch Linux package

rstacruz:
Adding a link to https://aur.archlinux.org/packages/kmonad-bin. I can confirm that it works :)


slotThe:
Thanks!

PR #208, opened 2021-04-05
Add more compose sequences

### LINKED ISSUE #207: Some specific compose characters not working
I've been trying out the compose characters feature. It works great for the most part, but I came across a handful of characters which don't seem to be supported in kmonad, despite there being compose sequences available. 

Here are the characters I've noticed not working:
¢ 
≤ 
≥ 
↑ 
↓
(note ← → left/right arrows do work fine, it's only up/down that don't).

 [I'm talking about the printed arrow characters here, not the movement keys]

slotThe:
Closes #207


slotThe:
I'm going to merge this; have tested it locally and the symbols in question work for me now

PR #209, opened 2021-04-07
Add sticky keys

### LINKED ISSUE #202: Support for sticky modifier keys
I'm trying to figure out if this is supported or not. I've looked at the existing options, but none seem quite right as far as I can tell. What I'm after is being able to implement sticky modifiers, with a timeout, which would obey the following rules:

1. Tapping a key activates a modifier and keeps it active for a predefined timeout period.
2. Tapping another key during the timeout period causes the modifier to be applied to that key.
3. The modifier is deactivated after the first output (i.e. tapping two keys within the timeout period means the modifier only applies only to the first).
4. Tapping a key after the timeout is equivalent to not having typed the modifier key at all.

So, for example, let's say I want "sticky shift":

tap LSFT, tap a   -->  A
press LSFT, tap a, release LSFT  -->  A
tap LSFT, (wait), tap a  -->  a
tap LSFT, tap a, tap b   --> Ab

slotThe:
This adds very simple sticky keys.  These are keys that will behave as if they were pressed after just tapping them.  This behaviour wears off after the next button is pressed, which makes them ideal for things like a quick control or shift.  For example, tapping a sticky and then pressing `abc` will result in `Abc`.

The number after `sticky-key` is the timeout you want, in milliseconds.  If a key is tapped and that time has passed, it won't act like it's pressed down when we receive the next keypress.

One can create these keys with the `sticky-key` keyword:

``` clojure
  (defalias
    slc (sticky-key 500 lctl))
```

Closes #202 ... somewhat.  What it doesn't implement is a way to chain sticky keys, e.g. pressing sticky-ctrl and then sticky-sft should result in sticky-(ctrl-sft).  This would require a way to inspect the buttons that will be pressed when we receive a keycode, which seems (aside from weird global state hackery) to be not something that we can (easily) do with the current setup.

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -381,6 +381,40 @@
   md4 (around % b)               ;; BEWARE: %B, not %b, do you see why?
 )
 
+#| --------------------------------------------------------------------------
+                          Optional: sticky keys
+
+  KMonad also support so called "sticky keys".  These are keys that will
+  behave as if they were pressed after just tapping them.  This behaviour
+  wears off after the next button is pressed, which makes them ideal for
+  things like a quick control or shift.  For example, tapping a sticky and
+  then pressing `abc' will result in `Abc'.
+
+  You can create these keys with the `sticky-key' keyword:
+
+    (defalias
+      slc (sticky-key 500 lctl))
+
+  The number after `sticky-key' is the timeout you want, in milliseconds.  If
+  a key is tapped and that time has passed, it won't act like it's pressed
+  down when we receive the next keypress.
+
+  It is also possible to combine sticky keys.  For example, to
+  get a sticky shift+control you can do
+
+    (defalias
+      ssc (around
+           (sticky-key 500 lsft)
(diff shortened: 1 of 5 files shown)


slotThe:
I've been using this quite happily for the last couple of days; works like a charm for me.  

I would say we can merge this now so more people can give it a spin and think of a solution for the extended functionality mentioned in #202 some other time; @stevep99 what do you think?

stevep99:
Yep, I agree, it's working great for me too!

stevep99:
By the way, I found an interesting workaround - kind of - for combing modifier sticky keys...

Given:
```
(defalias 
  ls_s (sticky-key 1500 lsft)
  lc_s (sticky-key 1500 lctl)
  lsc_s (around @ls_s @lc_s)
)
```

The `@lsc_s` definition seems to successfully do a sticky shift and control with a single keypress, which I thought was a neat trick.

slotThe:
Oh that's really cool!  I added a note about this in the tutorial, thanks!

PR #212, opened 2021-04-12
Ignore SIGCHLD

### LINKED ISSUE #211: defunct procceses owned by kmonad when I run cmd-buttons
When I use cmd-button I seem to get defunct processes added to my ps list.

For example, setting cmd-button to run xinput and pressing the key will add a [xinput] defunct entry to my process list. Tutorial.kbd gives "[sh] defunct" entry.

Tested this in a miminal config and tried two kmonad versions. I've never seen defunct proccesses before so I'm not sure if I've done something wrong. Killing kmonad removes all the defunct processes from the ps list.

slotThe:
Closes #211

Don't watch the child process so it can be killed correctly upon
completion.

TODO: 
  - [ ] Figure out how to incorporate this (nicely; I'm hoping for no CPP here) into kmonad such that things don't break on systems other than GNU/Linux (which presumably happens right now with this patch).

### DIFF
--- a/src/KMonad/App.hs
+++ b/src/KMonad/App.hs
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE CPP #-}
 {-|
 Module      : KMonad.App
 Description : The central app-loop of KMonad
@@ -18,8 +19,11 @@ where
 
 import KMonad.Prelude
 
-import UnliftIO.Process (CreateProcess(close_fds), createProcess_, shell)
 import RIO.Text (unpack)
+#ifdef linux_HOST_OS
+import System.Posix.Signals (Handler(Ignore), installHandler, sigCHLD)
+#endif
+import UnliftIO.Process (CreateProcess(close_fds), createProcess_, shell)
 
 import KMonad.Action
 import KMonad.Button
@@ -66,7 +70,7 @@ makeClassy ''AppCfg
 data AppEnv = AppEnv
   { -- Stored copy of cfg
     _keAppCfg   :: AppCfg
-   
+
     -- General IO
   , _keLogFunc  :: LogFunc
   , _keySink    :: KeySink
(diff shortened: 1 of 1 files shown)


recallthis:
Hi, sure. Tested with my clean config but I'm still getting the same issue here.

In case I made an error (git novice), I downloaded from your github page, did git checkout -b new-session, checked if the changes were applied to source then compiled with stack install, and ran ./kmonad from the build directory. 

* https://github.com/slotThe/kmonad/tree/new-session.

recallthis:
Great stuff, tried it and seems to work with both configs. :)

slotThe:
@MaxGyver83 May I abuse you as the Windows person again?  Does this even compile on Windows?

As far as I can tell, we don't currently have a "setup" step where I could plant this and I don't think it's necessarily worth creating a new abstraction just for this, so `CPP` it is, I guess...

MaxGyver83:
@slotThe : Did I compile KMonad on Windows before? I can't remember. Yes, I can try. Hopefully tomorrow!

MaxGyver83:
It looks like I have never compiled KMonad on Windows.

The KMonad README says I need a [Haskell platform installation](https://www.haskell.org/platform/). They say, I first need to install chocolatey, then `haskell-dev` via chocolatey. This worked so far. But then I couldn't find `stack` anywhere. So I tried `choco install stack`. This fails with a checksum error. 😞

nightscape:
@MaxGyver83 just a generic and maybe not helpful tip, but in general the package quality in Scoop is much better than in Chocolatey.

slotThe:
> It looks like I have never compiled KMonad on Windows.

Ah, no worries then!  I'll just add some CPP, we can fix this later if there's any better way I'm not seeing right now

david-janssen:
@MaxGyver83 

Could you add that to our actual documentation and make a pull request? Would be greatly appreciated :)

MaxGyver83:
Sure! Here it is: #238

david-janssen:
I...
hate...
Windows...
soooooooooooooooooooo MUCH!

*breathe breathe breathe breathe*

Couldn't get your instructions to work. Turns out I needed to *open the "haskell.downloads.org" website, not in Edge, not in Chrome, in Internet Explorer. Just so some certificate would be written to the right place, so that stack could interface with TLS...

Windows really is like witchcraft. I think for the next step in compilation I'm going to need some goatsblood.

david-janssen:
So, after also figuring out that I can't compile inside a VM-ware shared folder, your instructions were absolutely spot on. Thanks Max.

PR #191, opened 2021-02-23
Rebase develop onto master

thoelze1:
People are having issues working on `develop` because it lacks bug fixes from `master`, so I rebased

Issue #135, opened 2020-11-07
[Windows] Issue with ' and \ keys on deflayer with UK QWERTY host layout


szabog:
**OS:** Windows 10 x64 v2004 (Build 19041)
**KMonad version:** 0.4.1 (kmonad-0.4.1-win.exe from the latest release)
**Keyboard form factor:** ANSI

Hey,

I've encountered an error in KMonad while trying to create a keyboard layout using UK QWERTY as the host layout on Windows. The issue is specifically with remapping the `'` and `\` keys. I can't identify the root cause of the problem other than KMonad doesn't like translating from UK QWERTY to my `deflayer` layout. The issue cannot be replicated on US QWERTY, not even if it's set under a different locale: I tested it with Polish – Polish (Programmers) QWERTY (the first layer of the keyboard layout is identical to US QWERTY).

### Steps to replicate:

1. Set Windows keyboard layout to UK QWERTY.

2. Launch KMonad with my keyboard layout file.

3. Realise the top row (below the numbers' row) and the home row characters don't output what is configured in the `deflayer` section:

* Top row: `q w f p g [ j l u y ; # o`

* Home row: `a r s t d ] h n e i '`

### Expected behaviour:

The keys remap as defined in the `deflayer` section:

* Top row becomes `q w f p g [ j l u y ; ' \`

* Home row becomes `a r s t d ] h n e i o`

### KMonad config file:

```
(defcfg
    input  (low-level-hook)
    output (send-event-sink)

    allow-cmd false
    fallthrough true
)

(defsrc
    esc     f1  f2  f3  f4  f5  f6  f7  f8  f9  f10 f11 f12
    grv     1   2   3   4   5   6   7   8   9   0   -   =   bspc
    tab     q   w   e   r   t   y   u   i   o   p   [   ]   \
    caps    a   s   d   f   g   h   j   k   l   ;   '   ret
    lsft    z   x   c   v   b   n   m   ,   .   /   rsft
    lctl lmet lalt           spc            ralt rmet cmp rctl
)

(deflayer colemak-wide
    esc     f1  f2  f3  f4  f5  f6  f7  f8  f9  f10 f11 f12
    grv     1   2   3   4   5   6   =   7   8   9   0   -   bspc
    tab     q   w   f   p   g   [   j   l   u   y   ;   '   \\
    caps    a   r   s   t   d   ]   h   n   e   i   o   ret
    lsft    z   x   c   v   b   /   k   m   ,   .   rsft
[...]

PR #226, opened 2021-05-10
Unification

david-janssen:
Continuation of #191, but instead of rebasing we are merging.

I have no idea what I am doing, gitwise, but this is a merge of develop into master. There are currently merging conflicts.

Ideally I would just have my own branch of master on my own fork of kmonad, but my fork of kmonad is the official one, which means I needed a branch which I've called master-merged. It will contain the 'unified' master and develop. Once it's down, we'll merge it into master and delete both 'master-merged' and 'develop'.

I am not sure if this is the best way of doing things, but sometimes a-way-that-might-work is better than the-best-way-i-do-not-know-yet.


david-janssen:
So, to unify develop and master I need to fix the keycode representations. This has broken mac-support. I have started adding in code that deals with these issues, but we are not there yet. When I have brought this branch up to the point where keycode has been well-done I will put a notification in this PR and boop the people involved so we can get mac-support back up.

molleweide:
Allright!

MaxGyver83:
> So, to unify develop and master I need to fix the keycode representations. This has broken mac-support. I have started adding in code that deals with these issues, but we are not there yet. When I have brought this branch up to the point where keycode has been well-done I will put a notification in this PR and boop the people involved so we can get mac-support back up.

Does this by chance also address this issue? https://github.com/david-janssen/kmonad/issues/135#issuecomment-741809884

thoelze1:
I was able to rebase develop onto master without fixing keycode representations and without breaking Mac support (see [this comment](https://github.com/david-janssen/kmonad/pull/191#issuecomment-784640771)). I'm going to reattempt this shortly (merging instead of rebasing) and make a PR. Then we will be back to a single branch at which point you can make a feature branch and add your keycode refactoring commits, and I can concurrently make a branch and add my commits supporting the latest Mac version.

david-janssen:
@MaxGyver83 It doesn't address it just yet. First we get to 1 branch. Then the keycode refactor will move a little bit of the keycode management responsibilities to the OSes, at which point it should be easier to address #135 in some windows-specific code.

@thoelze1 Ahhh, ok. I didn't feel confident with the Mac-stuff. Thanks for doing this :-)

@bard Thanks for the advice :-) I think we have something of a solution moving now, which I will keep moving forward with, but I will keep this in mind for the future. I think the core issue is that my git knowledge doesn't go much further than pull commit push and a bit of merge. Even rebase I start thinking.. Que?

PR #240, opened 2021-05-13
Add build instructions for Windows (using scoop) 

MaxGyver83:
Same PR as some minutes ago but now against `kmonad/kmonad` with an updated repo in the `git clone` command.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -69,6 +69,27 @@ installation](https://www.haskell.org/platform). I also needed to install
 [mingw](http://mingw.org) to provide `gcc`. With both the Haskell platform and
 `mingw` building `kmonad` under Windows10 should as simple as `stack build`.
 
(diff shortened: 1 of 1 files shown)


slotThe:
The ownership transfering here is remarkably smart, so your old PR is already live (14c6ea674225d0596f800b589c407803fed7d110)!

The URL still has to be changed, so I suppose you can use this PR to just do that so it's not wasted

MaxGyver83:
OK, then I don't know why GitHub told me, I was **two** commits ahead. I assumed the first commit didn't make it into `kmonad/kmonad`, only into `david-janssen/kmonad`. When I resolve this one, is it okay to commit the merge into master? Otherwise, they suppose creating a new branch.

MaxGyver83:
Thanks for resolving the conflict, David!

david-janssen:
My pleasure, it reminded me to check the documentation further, and there were actually loads of references to the old URL, so I fixed those too.

PR #257, opened 2021-05-17
implemented tap-macro-release

vherrmann:
@6gk wanted a feature that allows him to tap a button on keypress and another on keyrelease. So i made a small patch for him with a new button called `tap-macro-release` which works just like `tap-macro`, but in contrast it taps the last button on keyrelease. I don't know if this is actually needed in kmonad or if it wouldn't bloat it, but i'll just drop the patch off.

### DIFF
--- a/src/KMonad/App/Parser/TokenJoiner.hs
+++ b/src/KMonad/App/Parser/TokenJoiner.hs
@@ -352,8 +352,8 @@ joinButton ns als =
     -- Various compound buttons
     KComposeSeq bs     -> view cmpKey >>= \c -> jst $ tapMacro . (c:) <$> mapM go bs
     KTapMacro bs       -> jst $ tapMacro           <$> mapM go bs
-    KTapMacroRelease bs
-      -> jst $ tapMacroRelease            <$> mapM go bs
+    KTapMacroRelease bs ->
+      jst $ tapMacroRelease            <$> mapM go bs
     KAround o i        -> jst $ around             <$> go o <*> go i
     KTapNext t h       -> jst $ tapNext            <$> go t <*> go h
     KTapHold s t h     -> jst $ tapHold (fi s)     <$> go t <*> go h
--- a/src/KMonad/App/Parser/TokenJoiner.hs
+++ b/src/KMonad/App/Parser/TokenJoiner.hs
@@ -352,6 +352,8 @@ joinButton ns als =
     -- Various compound buttons
     KComposeSeq bs     -> view cmpKey >>= \c -> jst $ tapMacro . (c:) <$> mapM go bs
     KTapMacro bs       -> jst $ tapMacro           <$> mapM go bs
+    KTapMacroRelease bs
+      -> jst $ tapMacroRelease            <$> mapM go bs
     KAround o i        -> jst $ around             <$> go o <*> go i
     KTapNext t h       -> jst $ tapNext            <$> go t <*> go h
     KTapHold s t h     -> jst $ tapHold (fi s)     <$> go t <*> go h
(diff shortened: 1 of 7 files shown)


vherrmann:
Wait, i just realized that i forgot to fix the documentation :S

molleweide:
oh nice! I'm gonna study this and see if I can understand the code.

src/KMonad/App/Parser/TokenJoiner.hs
```diff
@@ -352,6 +352,8 @@ joinButton ns als =
     -- Various compound buttons
     KComposeSeq bs     -> view cmpKey >>= \c -> jst $ tapMacro . (c:) <$> mapM go bs
     KTapMacro bs       -> jst $ tapMacro           <$> mapM go bs
+    KTapMacroRelease bs
+      -> jst $ tapMacroRelease            <$> mapM go bs
```
slotThe, line 356: In order to be consistent with the rest of the code base, format this as either

``` haskell
    KTapMacroRelease bs-> jst $ tapMacroRelease            <$> mapM go bs
```
or
``` haskell
    KTapMacroRelease bs ->
      jst $ tapMacroRelease <$> mapM go bs
```

src/KMonad/Args/Joiner.hs
```diff
@@ -364,6 +364,8 @@ joinButton ns als =
                              c   <- view cmpKey
                              jst $ tapMacro . (c:) <$> isps bs csd
     KTapMacro bs mbD   -> jst $ tapMacro           <$> isps bs mbD
+    KTapMacroRelease bs mbD
+      -> jst $ tapMacroRelease           <$> isps bs mbD
```
slotThe, line 368:
```suggestion
    KTapMacroRelease bs mbD -> 
      jst $ tapMacroRelease <$> isps bs mbD
```

src/KMonad/Model/Button.hs
```diff
@@ -359,6 +362,15 @@ tapMacro bs = onPress $ go bs
     go (b:rst) = tap b >> go rst
```
slotThe, line 364:
```suggestion
```

slotThe (commented):
Generally, I think this should be implemented as an option to `tap-macro` instead of its own button.  In the style of #152, perhaps something like a `:release` keyword?

Either way, I think we should wait with this until the keycode refactor is through.  It seems that things are already underway with `KMonad.App.Parser` being added in 297e96261d9158951251ec8285d8a9cfe9e539ef.  It seems like that is soon to replace `KMonad.Args`, but the latter is simply not removed yet.  I suppose @david-janssen will know more :)

david-janssen:
So, I actually agree with a lot of @slotThe's comments about layouting and how this maybe should be an option instead of its own button. However, I merged this as is because:
A. The code is really ugly right now, because I am also just flying through like a hurricane. Once all the new stuff is in and all the old stuff is gone, I'm going to go through and make sure everything is tidy and documented. I'd feel like a hypocrite insisting on clean, well-documented stuff right now, because half the stuff I write is ugly.
B. I'm actually hoping to really redo the whole way we specify buttons, so although this button is *very* similar to `tap-macro`, and I can't really see a use-case difference, I think in a few months the whole way we specify buttons might be redone anyways, so I find it hard to be very insistent on parsimony.

So, congrats on your Haskell contribution! :-)

david-janssen:
> It seems like that is soon to replace KMonad.Args, but the latter is simply not removed yet. I suppose @david-janssen will know more :)

Well, not really remove, just move. Changing a little bit at the fringes of those modules, bup reorganizing a lot of our code into some more distinct categories. After the move to the organization I thought I should just keep commiting non-breaking changes, but then I realized it's better to just work on my own tinkering branch, and merge it all in when done. So that was one of the last changes I made to master. I've been tinkering over here: https://github.com/david-janssen/kmonad/tree/tinker

PR #262, opened 2021-05-21
Add note about Vim editor support

Hugo-Heagren:
This exists in the organisation but wasn't documented in the README.

See also #69.


david-janssen:
Thank you very much :)

PR #268, opened 2021-05-25
My kmonad configuration

salahdin-ahmed:
Finally, I feel like my configuration is polished enough to be published online for actual use. I think it may be useful for someone out there. I added a `README` to make things easier for people.
Thanks for the great program btw.


david-janssen:
Merged. Although we will be moving user keymaps to their own repo sometime in the not too distant future.

joshskidmore:
The vim keys are quite brilliant!

Issue #210, opened 2021-04-08
Create reference for all available functions


slotThe:
So far, all of the functions one can use to change key behaviour (`around`, `tap-hold-next-release`, ...) are mentioned in the [tutorial](https://github.com/david-janssen/kmonad/blob/master/keymap/tutorial.kbd).  However, it may be nice to have a sort "quick reference guide" for all functions and what they do.  This would improve discoverability considerably I think.

salahdin-ahmed:
I think making them into a man page would be very useful. However, I think it won't be so convenient for Windows users!

david-janssen:
I like the idea, but instead of a `man` page the central point of this documentation should be on Github somewhere. Then, if someone wants to turn this into a `man` page, be my guest.

salahdin-ahmed:
I think this issue should be closed after b0d4bec44be459f8c7b052fe7e9a6bd09413ce34 and 22a79f4e3658fbe3f672e3696a021104afaac221.

PR #270, opened 2021-05-27
Add quick reference guide

salahdin-ahmed:
Hello,

I made a quick reference guide because of #210. It is a kbd file like the tutorial and included examples of every mentioned function. I am not experienced with documentation stuff but I hope you find this useful.

Thanks for the time,

SalahDin

### DIFF
--- a/doc/quick-reference.md
+++ b/doc/quick-reference.md
@@ -1,0 +1,205 @@
+# Introduction
+Welcome to the KMonad quick reference guide. This document aims to provide a
+simple definition and a usable example for all possible configuration options in
+the KMonad program to act as a quick reference that you look into when forgetting
+a function or/and wanting to know something quickly. Notice that this document
+does not try to be a full guide or a detailed documentation, this is the aim of
+[tutorial.kbd](https://github.com/kmonad/kmonad/blob/master/keymap/tutorial.kbd).
+
+# Define Configuration
+
++ `defcfg`: configuration rules (input and output devices, etc).
+
++ `input`: define the input keyboard which the program will captures.  
++ `output`: define the output keyboard which the system will rely on.
+
+	- For Linux:
+       ```clojure
+       input  (device-file "/dev/input/by-id/usb-04d9_daskeyboard-event-kbd")
+       output (uinput-sink "My KMonad output")
+       ```
+	- For Windows:
+       ```clojure
+       input  (low-level-hook)
+       output (send-event-sink)
+       ```
+	- For MacOS:
+		```clojure
+		input  (iokit-name "my-keyboard-product-string")
+		output (kext)
+		```
++ `fallthrough` (boolean, defaults to `false`): allow all keys, even if not defined
+in defsrc.
+
++ `allow-cmd` (boolean, defaults to `false`): allow arbitrary shell commands to
+  be run as keys.  
+NOTE: This is very dangerous since something like binding `rm -rf /*` to a key
+and executing it on a key press is possible.
+
++ `cmp-seq` (key, defaults to `RightAlt`): compose key for Unicode input (Linux X11
+specific).
+
+	+ `cmp-seq-delay` (number): delay between each pressed key in a compose-key
+sequence
+```clojure
+(defcfg
+  ;; For Linux
+  input  (device-file "/dev/input/by-id/usb-04d9_daskeyboard-event-kbd")
+  output (uinput-sink "My KMonad output" "/run/current-system/sw/bin/sleep 1 && /run/current-system/sw/bin/setxkbmap -option compose:ralt")
+
+  cmp-seq ralt
+  cmp-seq-delay 5
+
+  fallthrough true
+
+  allow-cmd true
+)
+```
+## Layers, aliases and buttons
+
++ `defsrc`: define physical keyboard input keys, although its layout is very
+  similar to a layer, it is not a one and will not be used as a one.
+```clojure
+(defsrc
+  grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc
+  tab  q    w    e    r    t    y    u    i    o    p    [    ]    \
+  caps a    s    d    f    g    h    j    k    l    ;    '    ret
+  lsft z    x    c    v    b    n    m    ,    .    /    rsft
+  lctl lmet lalt           spc            ralt rmet cmp  rctl
+)
+```
+
++ `defalias`: define a button, can be referenced later using `@` in `deflayer`
+```clojure
+(defalias
+  num  (layer-toggle numbers) ;; Bind num to a button that switches to a layer
+  kil  C-A-del                ;; Bind kil to a button that Ctrl+Alt+deletes
+)
+```
+
++ `deflayer`: define as many layers as you please (to toggle, add or remove).
+```clojure
+(deflayer qwerty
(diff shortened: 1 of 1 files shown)


slotThe:
This is cool, thanks!  

What I had in mind for #210 specifically was perhaps a markdown file (or some similar markup language) such that we could export this to a manpage with pandoc if needed.  I probably wasn't too clear in the issue, so sorry about that; do you think you could change the .kbd file to something like that?

salahdin-ahmed:
Oh, sorry for the misunderstanding, I will rewrite it in markdown.

david-janssen:
Just popping in to say: this looks great. Well done!

salahdin-ahmed:
Hello,

So after I rewrote the kbd file in markdown and used `pandoc` to convert it into man page the result was not satisfying, I tried playing with the file a little bit but it looks like it is beyond my skills and it will take some time from me to make it more readable. 

Thanks for the time,

SalahDin

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+  output (uinput-sink "My KMonad output" "/run/current-system/sw/bin/sleep 1 && /run/current-system/sw/bin/setxkbmap -option compose:ralt")
+
+  cmp-seq ralt
+  cmp-seq-delay 5
+
+  fallthrough true
+
+  allow-cmd true
+)
+```
```
slotThe, line 43: Is this now an example of the whole thing up to this point?  I would preface this by a few lines of text, this is kind of hard to read as it stands

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+
++ `defsrc`: define layer of input keys
```
slotThe, line 45: Might be worth mentioning that this should reflect the physical keyboard and is _not_ going to be used as any layer

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+```
+(defalias ken (layer-next next-layer))
+```
++ `layer-switch`: change the base layer
+```
+(defalias kes (layer-switch switch-layer))
+```
++ `layer-add`: overlays a layer on top of the current layer  
++ `layer-rem`: removes the overlayed layer from the current layer  
+**NOTE**: ***Using them is very dangerous, it is easy to make configurations which breaks the keyboard!***
```
slotThe, line 102: This is perhaps a bit over the top, though the message is worth communicating.  I don't think we have to make it bold though :)

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+### Modded buttons
+`C-` : `(around lctl X)`  
+`A-` : `(around lalt X)`  
+`M-` : `(around lmet X)`  
+`S-` : `(around lsft X)`  
+Then `RC-`, `RA-`, `RM-`, and `RS-` behave exactly the same, except using the
+right-modifier.
+```
+(defalias Ca C-a) ;; this is equivalent to (defalias Ca (around Ctl a))
+```
```
slotThe, line 122: This seems a bit out of place here

doc/kmonad.kbd.5
```diff
@@ -0,0 +1,299 @@
+.\" Automatically generated by Pandoc 2.11.4
```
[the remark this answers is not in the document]
salahdin-ahmed: It is under the file format section since this is not about how to use kmonad as a cmd-line tool rather about how to **format a config file**, so I thought it would be more suitable there. If you think it should not be there it is totally okay.

I agree it is better to remove this file for now and tackle it later as a separate PR.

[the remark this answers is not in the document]
slotThe: Ah I see.  I was tripped up by it still being called `kmonad`.  When we later tackle this again we should remember to call it `kmonad-config` or something like that and have the `kmonad` page be the general commands one

doc/quick-reference-guide.md
salahdin-ahmed: Why? I thought it would be helpful to add this part. I do not see why this should be removed.

doc/quick-reference-guide.md
salahdin-ahmed: I thought that adding an example at the end of how the whole block should look like would be helpful. People may think they are separated parts and I think using an example is easier and clearer than trying to put the idea into words.

doc/kmonad.kbd.5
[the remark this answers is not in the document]
salahdin-ahmed: Resolved in 722615de83518045136f4fe0b9a7ce608cacac6c

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
++ `defcfg`: configuration rules (input and output devices, etc).
+
++ `input` and `output`:
+	- For Linux:
+       ```
```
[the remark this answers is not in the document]
salahdin-ahmed: Resolved in c2c2189

doc/quick-reference-guide.md
salahdin-ahmed: Resolved in 55671fd

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+	- For MacOS:
+		```
+		input  (iokit-name "my-keyboard-product-string")
+		output (kext)
+		```
++ `fallthrough` (boolean, defaults to `false`): allow all keys, even if not defined
+in defsrc.
+
++ `allow-cmd` (boolean, defaults to `false`): allow commands to be run as keys (very
+dangerous).
```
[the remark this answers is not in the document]
salahdin-ahmed: Resolved in 7dae6cc

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
++ `defcfg`: configuration rules (input and output devices, etc).
```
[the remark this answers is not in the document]
salahdin-ahmed: Resolved in 1966b53

salahdin-ahmed:
I added some sections because I felt the document was so much a mess. I would like to hear your opinion about it.

doc/quick-reference-guide.md
salahdin-ahmed: Resolved in 3c0f1d7

doc/quick-reference-guide.md
```diff
@@ -0,0 +1,185 @@
...
+
++ `input` and `output`:
```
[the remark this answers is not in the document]
salahdin-ahmed: Resolved in 8fd289d

slotThe (approved):
I think this looks nice for a start now, thanks a lot!

Could you rename the file to `quick-reference.md` and squash the commits?  Then we can merge

molleweide:
> I think this looks nice for a start now, thanks a lot!
> 
> 
> 
> Could you rename the file to `quick-reference.md` and squash the commits?  Then we can merge 

Which git command is used to squash commits?

salahdin-ahmed:
> > I think this looks nice for a start now, thanks a lot!
> > Could you rename the file to `quick-reference.md` and squash the commits?  Then we can merge
> 
> Which git command is used to squash commits?

[Combining Git commits with squash - YouTube](https://www.youtube.com/watch?v=V5KrD7CmO4o)

slotThe:
Thanks!

PR #293, opened 2021-06-26
Fix the NixOS SystemD service

flexagoon:
Add a delay before the service starts to ensure that all connected keyboards are initialized by the system

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -86,6 +86,7 @@ with lib;
             description = "KMonad Instance for: " +conf-name;
             serviceConfig = {
               Type = "simple";
+              Restart = "always";
               ExecStart =
                 "${cfg.package}/bin/kmonad ${kbd-path}" +
                   # kmonad will error on initialization for any unplugged keyboards


nix/nixos-module.nix
```diff
@@ -86,6 +86,7 @@ with lib;
             description = "KMonad Instance for: " +conf-name;
             serviceConfig = {
               Type = "simple";
+              ExecStartPre = "/run/current-system/sw/bin/sleep 10";
```
salahdin-ahmed, line 89: I can not test this by myself, since I am not using NixOS or even a systemd distro, but isn't a 10 seconds of sleep a lot? I think further testing to minimize this number should be done.

flexagoon: For me it's just enough to fully boot and enter my password. But I'll try making the delay shorter and tell you the results

flexagoon: Btw the delay has no effect on the system startup time tho since systemd starts services asynchronously, the only thing that this affects is how soon kmonad will take effect after turning on your computer

flexagoon: @salahdin-ahmed I minimized the delay to 4 seconds, this is the lowest that worked for me.

slotThe:
This seems like a hack; I'm convinced nix has a better way to deal with things like this.

@stites @WhatisRT would either of you care to chime in here?

flexagoon:
> This seems like a hack; I'm convinced nix has a better way to deal with things like this.
> 
> @stites @WhatisRT would either of you care to chime in here?

I think it's an issue with systemd starting the service too early, not with Nix

WhatisRT:
I agree with all of the above. I haven't tried this, but there's a `keys` systemd target on NixOS, maybe putting that in the requirements solves the issue? I'm not too familiar with systemd, but I'd think that there is some target that ensures the hardware is already initialized.

On a related note, maybe there should be a (non NixOS) systemd unit (maybe as a template) in the repository? Then issues like this would not be directly related to the NixOS module, and the only issue would be to keep those two in sync.

flexagoon:
I tried making `run-keys.mount` unit a requirement for the systemd service, but it didn't work

slotThe:
On Sat, Jun 26 2021 12:23, WhatisRT wrote:
> On a related note, maybe there should be a (non NixOS) systemd unit
> (maybe as a template) in the repository? Then issues like this would
> not be directly related to the NixOS module, and the only issue would
> be to keep those two in sync.

There is one, actually: startup/kmonad.service.  I think it would be
great of we could somehow use this in the nix module!

flexagoon:
@slotThe I checked `startup/kmonad.service`  and turns out it has an option `Restart=always`, so the service fails, but restarts automatically. I can add this to the nixos module, however, I can't really see why it would be better than just adding a delay as I initially suggested.

flexagoon:
Actually, the delay-based approach failed on my system right now. It happened for the first time, but the fact that it did happen suggests that the restart-based approach is more reliable

slotThe:
I have to say I really like @WhatisRT's suggestion of integrating the service that we have with the nix module; do you want to try doing this in this PR?  If not, we can merge this fix now and perhaps tackle that at a later time.

flexagoon:
Unfortunately I'm not really sure how to do that, so I think you can just merge this PR

slotThe:
Thanks!

Issue #264, opened 2021-05-22
continuous delivery of kmonad executable


elpddev:
### Usecase

I'm on a Ubuntu machine and the available kmonad executable is a bit dated (v0.4.1 12-sep-2020).

### Current solution

I've download the repo and have done a local haskell build on the main branch. This takes around ~15 minutes.

### Suggestion

Setup github actions (or other CI) to build, create a release, and deploy the assets to the release commits to the main branch.

I could try to investigate  a working github actions setup if this issue is something that it is desired.

thoelze1:
The first build with `stack` takes a long time because all libraries have to be built, but subsequent builds will be much faster. There is talk of using cabal directly instead of stack (see #261). I don't have much knowledge of Haskell build systems but maybe cabal offers faculties that would reduce the time of even the first build

slotThe:
On Sat, May 22 2021 04:58, Eyal Lapid wrote:
> ### Suggestion
>
> Setup github actions (or other CI) to build, create a release, and
> deploy the assets to the release commits to the main branch.
>
> I could try to investigate a working github actions setup if this
> issue is something that it is desired.

I'm not convinced we need it; regular releases seem to be a better fit
for kmonad from my POV.  For people wanting more bleeding edge software
cloning and doing a `stack install' doesn't seem so much to ask.

Not _a lot_ has happened since September (though with the merging of
`develop` that's not quite true anymore).  A new release is on the
horizon though, as soon as #259 (which is a big refactor) is merged.

david-janssen:
It's something I have considered as well, but it would be my first time setting anything like that up. I'll leave this open because I want to investigate the upsides and downsides of this (are there any downsides?). Then, once the current refactor has settled a bit, I'd be happy to take a closer look at this. Therefore: marked as 'someday'

elpddev:
Thank you for the fast replies. 

I've played a little with a github actions a little and got some simple POC stack build and github pre-release creation with artifact uploading. It missing real version determination ,and maybe more specific cache and others. Maybe this will help if implemented someday.

https://github.com/elpddev/kmonad/blob/master/.github/workflows/build-prerelease.yml

david-janssen:
Thank you very much, Ill have a closer look when the time comes :)

pataquets:
FYI: Cross-linking my submitted `Dockerfile` in #295, which might come handy later. HTH.

pataquets:
[...]

PR #295, opened 2021-06-29
Add Dockerfile

pataquets:
I just discovered KMonad and found it awesome. Thank you very much for sharing this fine piece of software.

Since I wanted to try the latest master branch, I had to go for building. I'm on Ubuntu Bionic and I didn't feel like to install more software just to build it. So I went for building under Docker. I've created a `Dockerfile` based on Haskell 9 on Debian Buster which just adds source files and runs build instructions.

With just this file, I've been able to get a binary freshly built from source by just doing this in the `Dockerfile` directory:
``` shell
# Build a Docker image which will contain the binary.
$ docker build -t kmonad-builder .
# Spin up a Docker container from the built image, which will copy the
# built binary to a directory bind-mounted to the host's current dir.
$ docker run --rm -it -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
# Clean up build image.
$ docker rmi kmonad-builder
```
And... ta-da! I found in my current dir a `kmonad` binary ready to use (actually, I'm using this since a couple days ago without problems, so far). It builds also on your `keycode-refactor` branch, but I've not used that binary.

As an added bonus, you can build Docker images directly from public repo URLs, whithout even needing to clone (mine used as example here, in case you want to copy/paste for testing):

``` shell
$ docker build -t kmonad-builder github.com/pataquets/kmonad.git
```

Now, about the image:
First and foremost, since I was struggling to find an up-to-date binary for my system and I was able to get it easily using Docker, as I do with other software, I wanted to share it in case it can be of any help to others. You only need Docker installed, since the build happens isolated with all the build toolchain already installed in the building container. No other software needed in the host. This is way easier, tidier and safer than having to install build deps and build on your system. Looks like the built artifacts availability of the project could benefit of some love.

This `Dockerfile` can even be used in a hack/build/test development workflow (it's optimized to take the most advantage possible of Docker's build cache).

It also might come handy later when you get to do some CI, as I've seen in #264.

Otherwise, as it is now, it does not fits the 'use Docker to run the app' use-case as it's usually done in Docker. Hooking into host's uinput/udev/etc from inside a Docker container would be a first for me to sort out (I've accomplished using host's FUSE and disk/audio devices from Docker containers in the past, thou). Also, the built artifact weighs 3.85 Gb (!), but this can be solved, if necessary. But anyway, I feel it is a useful enough tool worth contributing, and also my way of giving back to this awesome software. I would be more than happy to submit a documentation PR, once we sort out if this is OK.

I can be of any further help, do not hesitate to ask. I wish you all the best of health.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM haskell:9-buster
+
+WORKDIR /usr/src/kmonad/
+
+COPY ./stack.yaml /usr/src/kmonad/
+COPY ./kmonad.cabal /usr/src/kmonad/
+RUN stack setup
+
+COPY ./ /usr/src/kmonad/
+RUN stack build
+
+RUN \
+  stack install && \
+  chmod --verbose +x /root/.local/bin/kmonad
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -4,6 +4,7 @@ Jump to
 - [Compilation](installation.md#compilation)
   - [Using nix](installation.md#using-nix)
   - [Using stack](installation.md#using-stack)
+  - [Using Docker](installation.md#using-docker)
   - [Windows environment](installation.md#windows-environment)
   - [macOS](installation.md#macos)
 - [Binaries](installation.md#binaries)
@@ -28,6 +29,32 @@ If you would like `stack` to automatically copy the binary to a folder on your
 `$PATH`, you can use:
 ```shell
 stack install # Builds *and* copies
+```
+
+### Using Docker
+If you have Docker installed, you can build `kmonad` from source without the need to install anything else on your system, since the build container will always have all the needed build tools and dependencies (currently Haskell 9 on Debian Buster).
+This is very convenient if no binaries are available and you want to try some other branch, you don't want to install build tools or they're not available for your OS, etc. You can even use the provided `Dockerfile` for development testing. As of now, the built image is not meant to *run* `kmonad`, just to build it.
+
+Just do this from the `Dockerfile` directory:
+``` shell
+# Build the Docker image which will contain the binary.
+docker build -t kmonad-builder .
+
+# Spin up an ephemeral Docker container from the built image, to just copy the
+# built binary to the host's current directory bind-mounted inside the
+# container at /host/.
+docker run --rm -it -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
+
+# Clean up build image, since it is no longer needed.
+docker rmi kmonad-builder
+```
+You will find a `kmonad` binary in your current directory.
+
+As an added bonus, with recent Docker versions you can build images straight
+from public repo URLs, whithout even needing to clone the repo.
+Do this as the build step (the first one) in the previous instructions:
+``` shell
+docker build -t kmonad-builder github.com/kmonad/kmonad.git
 ```
 
 ### Using `nix`


slotThe (commented):
Thanks!

I have absolutely no idea about Docker, but if this works for you I'd say we can merge it for now and then fix any arising issues afterwards.  

I think it would be great if you could add some more documentation somewhere.  For example, these commands

> ```
> # Build a Docker image which will contain the binary.
> $ docker build -t kmonad-builder .
> # Spin up a Docker container from the built image, which will copy the
> # built binary to a directory bind-mounted to the host's current dir.
> $ docker run --rm -it -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
> # Clean up build image.
> $ docker rmi kmonad-builder
> ```

don't seem so obvious to me :)

pataquets:
@slotThe : Sure! I'll add the docs. A couple questions:
* Most of build rationale/docs will be a shortened version of the issue summary.
* I'm assuming I should add docs to doc/Readme in the building section.
* Not sure, but I guess binary is statically linked.
* As for Docker usage, since I'm a long time user, I'm unsure about what's not obvious about it. Since you're playing the not-familiar-with-Docker user role, which is the main target here, would you mind in ellaborating what's not clear, so I can keep it in mind?

Thanks for the prompt review.

slotThe:
>     * Most of build rationale/docs will be a shortened version of the issue summary.

That's totally fine—don't have to write things twice when there's already quite a good guide written down somewhere

>     * I'm assuming I should add docs to doc/Readme in the building section.

Yes I would put it [around here](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#compilation) to the other "how to compile" sections. 

>     * Not sure, but I guess binary is statically linked.

It probably is, but it's most likely linked against glibc.  People running other libc's (like musl, for example) will probably be able to run the executable.  I'm not sure if this is worth mentioning though; I would imagine no one sees "Debian Buster Docker image" and expects that to be linked against musl. 

>     * As for Docker usage, since I'm a long time user, I'm unsure about what's not obvious about it. Since you're playing the not-familiar-with-Docker user role, which is the main target here, would you mind in ellaborating what's not clear, so I can keep it in mind?

I'm honestly not sure if I'm in a position to judge this.  I'm less the "new to Docker and wants to install kmonad with it" user and more the "has never used Docker and doesn't plan on doing so for the forseeable future" :)  But really, it was just what to actually execute to get the image to build things, i.e. especially this part:

> With just this file, I've been able to get a binary freshly built from source by just doing this in the Dockerfile directory:
>
> ``` console
> # Build a Docker image which will contain the binary.
> $ docker build -t kmonad-builder .
> # Spin up a Docker container from the built image, which will copy the
> # built binary to a directory bind-mounted to the host's current dir.
> $ docker run --rm -it -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
> # Clean up build image.
> $ docker rmi kmonad-builder
> ```
> 
> And... ta-da! I found in my current dir a kmonad binary ready to use (actually, I'm using this since a couple days ago without problems, so far). It builds also on your keycode-refactor branch, but I've not used that binary.
> 
> As an added bonus, you can build Docker images directly from public repo URLs, whithout even needing to clone (mine used as example here, in case you want to copy/paste for testing):
> 
> ``` console
> $ docker build -t kmonad-builder github.com/pataquets/kmonad.git
> ```

Telling users that this is really only for building and one then needs to copy over the executable would also be valuable I think.

pataquets:
@slotThe: Docs added, I hope they're good enough.

I've checked my Docker-built binary for linking:
```
$ ldd `which kmonad`                            
        linux-vdso.so.1 (0x00007ffceeead000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1ae0315000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f1ae010d000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f1adff0a000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f1adfd06000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1adfae7000)
        libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f1adf866000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1adf475000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1ae06b3000)
```
I've digged a bit to learn about static compilation for maybe using it here. I've seen that it can be done with `nix`, which could be installed for building. But, if I was given the choice, ideally it would be just one more Make/Cabal/etc target, which would be passed as a build argument using Docker's [`ARG`](https://docs.docker.com/engine/reference/builder/#arg) statement.

slotThe:
> @slotThe: Docs added, I hope they're good enough.

Looks fine to me!

> I've digged a bit to learn about static compilation for maybe using it here. I've seen that it can be done with `nix`, which could be installed for building. But, if I was given the choice, ideally it would be just one more Make/Cabal/etc target, which would be passed as a build argument using Docker's [`ARG`](https://docs.docker.com/engine/reference/builder/#arg) statement.

If you want to do this in this PR then feel free to.  Don't feel pressured though, I would be happy to merge it as-is and perhaps if someone want to add this later on they can do it.

pataquets:
I would like to do it in another PR when/if the need arises. I expect it to be quite involving to achieve, so it would be a separate issue on its own.
I'm about to lose bandwidth due to holidays and I also would like to watch for feedback in case someone reports.

slotThe:
Sure; thanks!

pataquets:
@slotThe: May I suggest you to merge/cherry pick this into `keyboard-refactor`i branch to enable others to easily test that branch. I've just tried and it built successfully and running it without noticing any problem.

I've been reviewing some issues and this might lower the friction for some interested parties who otherwise may not take the plunge to test it due to building complexity.

Issue #296, opened 2021-06-29
Add --version switch


pataquets:
While tinkering with building both `master` and `keycode-refactor` branches in #295, I've missed some kind of `--version` switch to check the binary version, ideally with the build commit short hash.
No idea on how easy this could be, just dropping the idea.

slotThe:
On Tue, Jun 29 2021 08:10, Alfonso Montero wrote:
> While tinkering with building both `master` and `keycode-refactor`
> branches in #295, I've missed some kind of `--version` switch to check
> the binary version, ideally with the build commit short hash.
>
> No idea on how easy this could be, just dropping the idea.

I like this idea :+1:

The version itself would be trivial, as cabal knows this and we can
query it through the generated `Paths_kmonad` module.

The git hash would be a little bit more complicated.  We'd probably need
to query that from git at compile via TemplateHaskell.  I can see that
TemplateHaskell is already in our `default-extensions` (although I do
not know where it is used), so perhaps that approach isn't even so
far-fetched.

pataquets:
@slotThe : If you have handy the info to easily do this, I would love to add it to #295, if it can be done at build time. Also, any other feedback on the issue is welcome :grin: .

slotThe:
> @slotThe : If you have handy the info to easily do this, I would love to add it to #295, if it can be done at build time.

This solution would be entirely Haskell-based, so the docker image woudn't have to change at all!

PR #301, opened 2021-07-03
Add `--version` (`-V`) flag

slotThe:
Adds a --version (-V) flag to the CLI.  It prints the current version as
noted in the cabal file, as well as (if possible) a hash of the revision
that kmonad was compiled with.  The bulk of the changes are just the
overhead of creating a new module for TemplateHaskell (GHC does not like
local definitions of this stuff).  Requires `git` to be in $PATH, though
is seems like a reasonable assumption.

Fixes: https://github.com/kmonad/kmonad/issues/296

@pataquets would you care to give this a try?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Added macOS 11.0 support
 - Added a `sticky-key`
 - Expanded documentation
+- Added `--version` (`-V`) flag
 
 ### [Changed]
 - Reorganized codebase
--- a/kmonad.cabal
+++ b/kmonad.cabal
@@ -47,6 +47,7 @@ library
     , rio
     , time
     , unliftio
+    , template-haskell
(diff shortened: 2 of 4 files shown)


pataquets:
Sure, @slotThe!

``` shell
$ ./kmonad --version
0.4.1, commit f4c8554b19638354c68ad0bed34c5fec19f706f8
$ ./kmonad -V
0.4.1, commit f4c8554b19638354c68ad0bed34c5fec19f706f8
```
Note: disregard commit-id, since I've had to merge in my branch from #295 to build.
As of now, I'm using this version on my PC.

slotThe:
@WhatisRT  @flexagoon would you care to give this a spin on NixOS?  I wonder if this compiles there or whether we also need to adjust these files

slotThe:
Okay I just got my hands on a nix machine and it succeeded building with both the `nix-shell` and the `nix-build` command that are referenced in the README.  Seems safe enough to merge this now

PR #302, opened 2021-07-05
Remove inconsistencies between a normal service and a NixOS one

flexagoon:
Kmonad systemd modules previously had an opition `Restart=always`. It was doing literally that - restarting the service every several seconds, even if it didn't produce any errors. This PR changes it to `Restart=on-failure`, so that the service only restarts in case it fails.

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -87,6 +87,8 @@ with lib;
             serviceConfig = {
               Type = "simple";
               Restart = "always";
+              RestartSec = 3;
+              Nice = -20;
               ExecStart =
                 "${cfg.package}/bin/kmonad ${kbd-path}" +
                   # kmonad will error on initialization for any unplugged keyboards


slotThe:
On Sun, Jul 04 2021 22:40, flexagoon wrote:
> Kmonad systemd modules previously had an opition `Restart=always`. It
> was doing literally that - restarting the service every several
> seconds, even if it didn&#39;t produce any errors.

I'm not a systemd user, but I believe that's wrong.  It restarts the
process _when it exists_ regardless of whether that exit what successful
or not[1]:

  > If set to always, the service will be restarted regardless of whether
  > it exited cleanly or not, got terminated abnormally by a signal, or
  > hit a timeout.

The kmonad process should not exit unexpectedly until you kill it; is
that not what you're experiencing?

[1]: https://man7.org/linux/man-pages/man5/systemd.service.5.html#OPTIONS

flexagoon:
Oh, sorry. Looks like I've misread the documentation. Nixos module was out-of-sync with the normal one, so I guess I can reuse this commit for that

PR #305, opened 2021-07-07
Cancel multi-tap when another key is pressed (closes #304)

### LINKED ISSUE #304: Cancel multi-tap when pressing another key
Hi. First, thanks for this nice piece of software.

I am still in the process of trying to find the best configuration to suit my needs.
Doing so i experimented with multi-tap and found something that might be improved.

So I tried something like:
```lisp
(defalias
    mymulti (multi-tap 180 1 2)
)
```
When the daemon is started with this configuration, if you type key 1 then key 3 in less than 180 ms, then the output is "3, 1" when I would expect "1, 3".

I understand why kmonad is doing this: it is still waiting for another tap on 1 before deciding whether it's a 1 or a 2. But if another key is pressed just after 1, it is clear the intent of the typist was not to repeat a tap on 1.

So my suggestion: wouldn't it make sense to cancel the multi-tap behavior as soon as anything happens between the two taps?

adegorre:
This PR will close #304. It changes the logic of multi-tap such that pressing another key while still in the multi-tap delay cancels the delay and just taps the current button of the multi-tap sequence, as described in #304.

There is still some issue with modifier keys: assuming `c` is the first symbol of the multitap sequence and I want to do `lctl+c`, it will not work when I release `lctl` before the delay.
(Probably this event was already eaten by the event manager associated to `lctl`.)

Any idea how to improve this?

(also this is my first Haskell code ever, please be gentle but I will gladly listen to constructive criticisms!)

### DIFF
--- a/src/KMonad/Model/Button.hs
+++ b/src/KMonad/Model/Button.hs
@@ -336,21 +336,32 @@ multiTap l bs = onPress $ go bs
     go []            = press l
     go ((ms, b):bs') = do
       -- This is a bit complicated. What we do is:
-      -- 1.  We wait for the release of the key that triggered this action
-      -- 2A. If it doesn't occur in the interval we press the button from the list
-      --     and we are done.
-      -- 2B. If we do detect the release, we must now keep waiting to detect another press.
-      -- 3A. If we do not detect a press before the interval is up, we know a tap occured,
-      --     so we tap the current button and we are done.
-      -- 3B. If we detect another press, then the user is descending into the buttons tied
-      --     to this multi-tap, so we recurse on the remaining buttons.
-      let onMatch t = do
-            within (ms - t^.elapsed) (matchMy Press)
-                   (tap b)
-                   (const $ go bs' *> pure Catch)
-            pure Catch
-      within ms (matchMy Release) (press b) onMatch
-
+      -- 1.  We wait for an event
+      -- 2A. If it doesn't occur in the interval we press the button from the
+      --     list and we are done.
+      -- 2B. If we do detect the release of the key that triggered this action,
+      --     we must now keep waiting to detect another press.
+      -- 2C. If we detect another (unrelated) press event we cancel the
+      --     remaining of the multi-tap sequence and trigger a tap on the
+      --     current button of the sequence.
+      -- 3A. After 2B, if we do not detect a press before the interval is up,
+      --     we know a tap occurred, so we tap the current button and we are
+      --     done.
+      -- 3B. If we detect another press of the same key, then the user is
+      --     descending into the buttons tied to this multi-tap, so we recurse
+      --     on the remaining buttons.
+      -- 3C. If we detect any other (unrelated) press event, then the multi-tap
+      --     sequence is cancelled like in 2C. We trigger a tap of the current
+      --     button of the sequence.
+      let doNext pred onTimeout next ms = tHookF InputHook ms onTimeout $ \t -> do
+            pr <- pred
+            if | pr (t^.event)      -> next (ms - t^.elapsed) $> Catch
+               | isPress (t^.event) -> tap b                  $> NoCatch
+               | otherwise          -> pure NoCatch
+      doNext (matchMy Release)
+             (press b)
+             (doNext (matchMy Press) (tap b) (\_ -> go bs'))
+             ms
 
 -- | Create a 'Button' that performs a series of taps on press. Note that the
 -- last button is only released when the tapMacro itself is released.
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 ### [Changed]
 - Reorganized codebase
+- The `multi-tap` key now immediately taps the current key when another
+  key is pressed during tapping.
 
 ### [Fixed]
 - Fixed compilation error under Mac, having to do with typo in Keycodes
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -789,7 +789,8 @@
   without delay. As long as you tap the `multi-tap` within the delay specified,
   it will jump to the next button. Once the delay is exceeded the selected
   button is pressed. If the last button in the list is reached, it is
-  immediately pressed.
+  immediately pressed. When another key is pressed down while we're tapping,
+  `multi-tap' also immediately exits and taps the current button.
 
   Note that you can actually hold the button, so in the below example, going:
   tap-tap-hold (wait 300ms) will get you a pressed c, until you release again.


adegorre:
Converted to a draft because of the modifier issue and because I noticed I should also update the comment that explains the logic.

adegorre:
Now I understand why foreign release events are ignored: simple keys already have their own registered hook that was registered when the key was pressed. It seems that the hook that was registered first takes priority and eats the event.

As this is obviously a more global problem (with other effects such as making it impossible to fix #307), and this PR already improves current behavior, there is no reason anymore to consider this PR as a draft.
Moreover if, some day in the future, foreign release events can finally be received, they should also be handled correctly by the logic of this PR.

src/KMonad/Model/Button.hs
```diff
@@ -336,21 +336,32 @@ multiTap l bs = onPress $ go bs
...
+                  tap b
+                  pure NoCatch
+          onMatchPress t = do
+            pp <- matchMy Press
+            if pp $ t^.event then do
+              go bs' *> pure Catch
+            else do
+                  tap b
+                  pure NoCatch 
+      within ms (pure (const True)) (press b) onMatchRelease
```
slotThe, line 364: Since we're matching on any event this may well just be a `tHookF` (ditto above)

src/KMonad/Model/Button.hs
```diff
@@ -336,21 +336,32 @@ multiTap l bs = onPress $ go bs
...
-                   (const $ go bs' *> pure Catch)
-            pure Catch
-      within ms (matchMy Release) (press b) onMatch
-
+      -- 3C. If we detect any other (unrelated) event, then the multi-tap sequence is cancelled like in 2C. We trigger a tap of the current button of the sequence.
+      let onMatchRelease t = do
+            pr <- matchMy Release
+            if pr $ t^.event then do
+              within  (ms - t^.elapsed) (pure (const True)) (tap b) onMatchPress
+              pure Catch
```
slotThe, line 353: I feel like most of these cases could be written more concisely in the style of `Catch <$ within ...`

slotThe (commented):
Thanks for the contribution! 

On the face of it, this sounds like a logical change.  However, I think we should exclude key releases because that could lead to undesired effects.  When typing fast, I would imagine it's not uncommon to release keys pressed before the multi-tap key; in this case, do we really want the key to be tapped every time?

src/KMonad/Model/Button.hs
adegorre: Got it. At the time I wrote this I did not know `within` and `tHookF` were basically the same mechanism.

src/KMonad/Model/Button.hs
adegorre: I converted most of my code so that it uses this more concise style.
I also factorized it a little further as to avoid code replication.

adegorre:
Thank you for your useful review!

I think I addressed all your concerns and suggestions. I hope current version is fully satisfying now!

---

> On the face of it, this sounds like a logical change. However, I think we should exclude key releases because that could lead to undesired effects. When typing fast, I would imagine it's not uncommon to release keys pressed before the multi-tap key; in this case, do we really want the key to be tapped every time?

Well actually, while using multi-tap, I already had problems when releasing the shift key too soon (before the delay) which resulted in printing the lower case letter when i expected an uppercase one.
So I assume in some cases (when the previous key was a modifier) one would prefer the release event to be a cancellation cause, while in some other cases (when the previous key was a regular alphabetical key) one would prefer the release event to be ignored by multi-tap.

Nonetheless, if the release event would cancel multi-tap, it would be a disrupting change for people used to the other behavior so you are right it should not be changed (but maybe an option or a multi-tap variant should be proposed).
So I made the code to ignore foreign release events, as you suggested.

(anyway this discussion is currently moot: foreign release events are not received by the multi-tap hook. If i understand Kmonad well enough, it seems that no release event can be handled by a foreign hook.)

src/KMonad/Model/Button.hs
```diff
@@ -336,21 +336,25 @@ multiTap l bs = onPress $ go bs
...
-                   (const $ go bs' *> pure Catch)
-            pure Catch
-      within ms (matchMy Release) (press b) onMatch
-
+      -- 3C. If we detect any other (unrelated) press event, then the multi-tap sequence is cancelled like in 2C. We trigger a tap of the current button of the sequence.
+
+      let doNext pred onTimeout next ms = tHookF InputHook ms onTimeout (\t -> do
+            pr <- pred
+            if 
+              | pr $ t^.event -> Catch <$ next (ms - t^.elapsed)
```
slotThe, line 353: Mh is this correct?  The old version called `within` twice in order to adjust the delay when we detected our own release—now we just recurse while ignoring that.

slotThe:
> Anyway all this discussion is currently moot: foreign release events are not received by the multi-tap hook. If i understand Kmonad well enough, it seems that no release event can be handled by a foreign hook.)

I'm not sure what you mean by "foreign" release events and hooks, but it's certainly true that we only get the trigger key here (i.e., the actual key on the keyboard that was pressed) and don't really have any way to find out whether it's mapped to a modifier key or not

adegorre:
> I'm not sure what you mean by "foreign" release events and hooks, but it's certainly true that we only get the trigger key here (i.e., the actual key on the keyboard that was pressed) and don't really have any way to find out whether it's mapped to a modifier key or not

By "foreign", I mean the release of another key. If I understand correctly, when a key is pressed, it immediately installs a hook waiting for its own release. This hook prevents other hooks from handling this release event. Is this correct?

src/KMonad/Model/Button.hs
adegorre: Isn't it what happens here?

`tHookF` is called twice (through the 2 calls to `doNext`), first with `ms` as an argument (the parameter of `go`), then with `ms - t^.elapsed` where `t` is the trigger of the first hook. So the delay is adjusted in the same way as it used to be, or do I get it wrong?

slotThe: I may be being stupid, but where is `doNext` called twice?  It's certainly called once with the `ms` argument, but `next` is then defined as `\_ -> go bs')`, meaning the new delay will simply be ignored.

adegorre: `doNext` is passed as a callback to itself: in the first call, line 357, `(doNext (matchMy Press) (tap b) (\_ -> go bs'))` is passed as parameter `next` (4 arguments out of 5: time is missing).

Hence at line 353, during the first call, `next (ms - t^.elapsed)` develops to `doNext (matchMy Press) (tap b) (\_ -> go bs') (ms - t^.elapsed)`, making the second call (with corrected delay).

I hope this is correct.

Ok, even assuming this is correct, code that makes readers feel stupid can't be very good code!
`doNext` could be made into 2 functions again like in the previous version if it makes it more readable...

slotThe: Oooh of course.  `doNext` has quite a lot of arguments and I must have mixed up some parentheses there.  Sorry about that

adegorre: No worry!

Please let me know if I should do something to improve readability or if there are some coding guidelines I should be aware of.

slotThe: I think this is really fine; perhaps a type signature on `doNext` could help with understanding.  Other than that I'll try to test this myself tomorrow and then we can probably merge

slotThe:
Thanks!

PR #308, opened 2021-07-09
Add missing nix dependency

dlip:
#301 added a git requirement to the build which is not available in the nix build environment. This fix is still probably not ideal as it results in an empty commit hash, but at least it fixes the build.

### DIFF
--- a/nix/kmonad.nix
+++ b/nix/kmonad.nix
@@ -1,6 +1,6 @@
 { mkDerivation, base, cereal, lens, megaparsec, mtl
 , optparse-applicative, resourcet, rio, stdenv, time, unix
-, unliftio
+, unliftio, pkgs
 }:
 mkDerivation {
   pname = "kmonad";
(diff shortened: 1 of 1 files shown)


vherrmann:
Are you sure about it resulting in an empty hash? I just built kmonad with `nix-build nix` and `kmonad --version` printed the right hash.

Btw: Why do you know/Where did you find the option `buildDepends`? I couldn't find it anywhere….

dlip:
I think it resulted in an empty hash for me because I'm including it from my flake, so maybe it didn't clone the full repo. Building from the cloned kmonad repo might work ok.

I found `buildDepends` here
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/generic-builder.nix#L18

Nix is not documented well in many areas so I usually go looking in the code

slotThe:
Oh so it _is_ possible, just undocumented...  That's always a lot of fun.  Thank you for digging!

Issue #314, opened 2021-07-20
Kmonad not grabbing keyboard


destroyer449:
I've started trying to use kmonad, and I've ran into a problem that I don't know how to solve. I wrote a fairly basic config for kmonad:
```(defcfg
  input  (device-file "/dev/input/event4")
  output (uinput-sink "Kmonad"
    "/usr/bin/sleep 1 && /usr/bin/setxkbmap -option compose:ralt")
  cmp-seq ralt    ;; Set the compose key to `RightAlt'
  cmp-seq-delay 5 ;; 5ms delay between each compose-key sequence press

  ;; Comment this is you want unhandled events not to be emitted
  fallthrough true

  ;; Set this to false to disable any command-execution in KMonad
  allow-cmd true
)

;;Default layer
(defsrc
    esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12  ssrq ins del home end pgup pgdn
    grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc         nlck kp/ kp*  kp-
    tab  q    w    e    r    t    y    u    i    o    p    [    ]    \            kp7  kp8 kp9  kp+
    esc  a    s    d    f    g    h    j    k    l    ;    '         ret          kp4  kp5 kp6
    lsft z    x    c    v    b    n    n    m    ,    .    /    rsft  up          kp1  kp2 kp3  kprt
    lctl lmet lalt          spc                  ralt rmet rctl left down  right  kp0  kp.)

#| Blank layer template
(deflayer name
    _    _    _    _    _    _    _    _    _    _    _    _    _    _    _   _   _    _   _    _
    _    _    _    _    _    _    _    _    _    _    _    _    _    _            _    _   _    _
    _    _    _    _    _    _    _    _    _    _    _    _    _    _            _    _   _    _
    _    _    _    _    _    _    _    _    _    _    _    _         _            _    _   _
    _    _    _    _    _    _    _    _    _    _    _    _    _    _            _    _   _    _
    _    _    _             _                    _    _    _    _    _     _      _    _  )
|#

(defalias cesc (tap-next esc lctl)
)

(deflayer default
    esc   f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12  ssrq ins del home end pgup pgdn
    grv   1    2    3    4    5    6    7    8    9    0    -    =    bspc         nlck kp/ kp*  kp-
[...]

PR #317, opened 2021-07-23
Fix error when executing command in defsrc output

MuhammedZakir:
Ref: https://github.com/kmonad/kmonad/issues/314#issuecomment-885129199


slotThe:
Thank you!

PR #325, opened 2021-08-06
Logitech G15 G11

Lucki:
This is a sample configuration for the extra G-Keys used on the original Logitech G15 and Logitech G11 with 18 G-Keys. (There are newer variants with less keys)

Running g15daemon / libg15 maps the G-Keys to some rarely used media keys which allows them to be used in programs or shortcut managers without having to run an additional userspace counterpart like gnome15 or the original Logitech profile software used to do but has the problem that the media keys may already get picked up by some desktop environments. That's why I've also assigned some modifier to every set of keys.

Speaking of sets there are originally only three sets (M1-M3) but without additional tinkering the macro record button (MR) would still be a dumb media key so I've assigned a fourth set to it.

Personally I've set the G-Keys to call a little script which looks for the currently active application and calls other scripts based on that and even allows for toggling scripts. Together with this config it basically restores the known behavior of 3 (+1) macro sets per profile (active application).

### DIFF
--- a/keymap/user/Lucki/logitech_g15_g11.kbd
+++ b/keymap/user/Lucki/logitech_g15_g11.kbd
@@ -1,0 +1,171 @@
+#|================= Logitech G15 / G11 running g15daemon =====================|#
+
+#|  This is the layout g15deamon / libg15 maps the G-Keys to on the original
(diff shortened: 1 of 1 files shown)


slotThe:
Thanks!

PR #329, opened 2021-08-09
Fix doc for user getting started.

jcguu95:
1. Fix typo.
2. `input` and `uinput` are both required.
3. `by-id` won't always work.


slotThe:
Thank you!

PR #338, opened 2021-08-19
Update nix derivation

terlar:
This fixes a few things in the nix derivation.

- Update the version to the same latest tagged version
- Replace deprecated usage of stdenv.lib, the lib is the preferred input.
- Improve the speed of fetching pinned nixpkgs by using tarball instead.


slotThe (approved):
LGTM!  I will merge this in a few days if no-one objects

Issue #335, opened 2021-08-15
Explicitly releasing modifier keys?


lierdakil:
I've been using kmonad (master branch) for a few days, and it feels great so far, but I've run into a bit of a limitation, either in kmonad or my own inventiveness :)

I want to swap symbols and numbers on my top number row, i.e. send shift+5 when key 5 is pressed and vice versa. It's easy enough to do the first one, i.e. send a modified key when an unmodified key is pressed. The opposite turns out to be much less straightforward. I could of course create a new layer wherein all keys are shifted except the number row, but that would effectively duplicate my config, which is not optimal to say the least.

The "simple" solution would be a way to explicitly send key release events, like a reverse of `around` function. However, I didn't find a way to achieve that.

Any suggestions?

Also, I'll ask here to avoid creating multiple issues. I've made a crude Dockerfile to build a static Linux executable (Alpine-based). It's easily convertible to GitHub actions. Would you be interested in adding automated builds to this repo that way? Fair warning: this will need a bit of help from someone familiar with stack; the Dockerfile uses cabal because that's what I'm more comfortable with, but I see no reason it couldn't be converted to use stack. I'll create a draft PR later this week if this sounds interesting.

slotThe:
As for the first question, this is another incarnation of #111.  The tl;dr is that you need to create a "shift layer" where everything is shifted up but the things you want.  The idea of simply sending a release event is enteresting though—I don't think it would break anything because kmonad simply ignores double releases.  I will try to look into this a bit further.

> I've made a crude Dockerfile to build a static Linux executable (Alpine-based). It's easily convertible to GitHub actions. Would you be interested in adding automated builds to this repo that way? Fair warning: this will need a bit of help from someone familiar with stack; the Dockerfile uses cabal because that's what I'm more comfortable with, but I see no reason it couldn't be converted to use stack. I'll create a draft PR later this week if this sounds interesting.

This sounds interesting!  We already have a dockerfile, so if at all possible it would be nice to incorporate that into the existing file

adgai19:
> As for the first question, this is another incarnation of #111. The tl;dr is that you need to create a "shift layer" where everything is shifted up but the things you want.
It would be nice if there was a config file or example available for this, as many programmer layouts do swap shift and non-shift keys of number row.

oati:
An `unaround` / `unwrap` function would need to
- on key down: release a specified key
- on key up: press the specified key if there is a currently-pressed physical key corresponding to the specified key
[...]

PR #340, opened 2021-08-20
Docker static build

lierdakil:
As discussed in #335, this is a Dockerfile for building a static executable. Additionally, I've added a GitHub action that builds the executable and uploads it as an artefact. I've figured out how to convert it to use stack instead of cabal.

I'm using my own docker image as the base, but it's simply an Alpine image with some dependencies and GHC installed via ghcup, the sources are at https://github.com/lierdakil/alpine-haskell. Note that at the time of writing, stack can't install newer GHC versions on Alpine, so we have to go through ghcup (or build from source)

I did basically overwrite the original Buster-based Dockerfile, as I understood that to be the intention stated in #335. Besides, building a non-static executable in a Docker environment has questionable utility if it's meant to run on the host, unless the host matches the Docker environment pretty closely. However, we can also have two Dockerfiles if that's preferable.

I've also fixed a typo in the docs with https://github.com/kmonad/kmonad/commit/f35c8cee959c43a76e0278a28a02b8b6b99e544a -- I don't really want to create a separate PR for one-character patch.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+.ghc*
+dist*
+.stack*
+configs*
+source_maps*
+stack.yaml.lock
+src_solid
+log.log
+tmp
+template/testing.kbd
+bin/
+result
+kmonad
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,32 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ '*' ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v2
+
+      - uses: actions/cache@v2
+        with:
+          path: image.tar
+          key: ${{ runner.os }}-primes
+
+      - run: |
+          [ -r image.tar ] && docker load -i image.tar || true
+          rm image.tar || true
+
+      - run: docker build . -t kmonad-builder --cache-from=kmonad-builder:latest
+
+      - run: docker run --rm -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: kmonad-binary
+          path: kmonad
+
+      - run: docker save kmonad-builder -o image.tar
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,14 @@
-FROM haskell:9-buster
+FROM lierdakil/alpine-haskell:8.10.4
 
 WORKDIR /usr/src/kmonad/
+RUN apk --no-cache add git
+RUN stack update
 
-COPY ./stack.yaml /usr/src/kmonad/
-COPY ./kmonad.cabal /usr/src/kmonad/
-RUN stack setup
-
-COPY ./ /usr/src/kmonad/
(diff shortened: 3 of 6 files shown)


slotThe (commented):
This looks great, thank you!

There is a section about docker in [installation.md](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#using-docker); could you check whether that's still applicable to the new file?

Also cc'ing @pataquets, who contributed the original dockerfile.

lierdakil:
@slotThe, everything is applicable, except the remark about the image being Debian-based. Maybe we can remove it altogether, since it's not really relevant to the installation process.

slotThe:
> @slotThe, everything is applicable, except the remark about the image being Debian-based. Maybe we can remove it altogether, since it's not really relevant to the installation process.

Sounds good to me :+1:

lierdakil:
Ok. I've amended the docs, and made a couple minor performance-related tweaks.

slotThe:
Perfect, thanks!

molleweide:
Should I expect the docker installer to work on macos?
After having run the docker installer I get `cannot execute binary file`.

lierdakil:
@molleweide, since docker runs a Linux image, it builds a Linux executable... so probably not, sorry.

pataquets:
@molleweide : Since now `kmonad`is built linking to `musl` instead of `libc`, you can give a try building the previous `glibc` build by doing:
```
docker build -t kmonad-builder github.com/kmonad/kmonad#fa407a6df7c428dd3f74b5fce7251359aa96d42f
```

I'm not sure if, being macOS an Apple-ish Linux (or at least *nix) flavor, it should compile using `glibc` inside a macOS Docker build. I can't tell for sure, but I won't rule it out. I'm not a macOS user, sorry.

Also cc'ing @slotThe : This makes me wonder if you're aware and confortable with `kmonad` linking to `musl` instead of `glibc`. Can't this cause hard-to-reproduce errors on developer's build environments?

pataquets:
I must say that I'm using `kmonad` version built with `musl` since a few weeks ago, for testing it, with no problems so far on my Linux box.

slotThe:
On Thu, Sep 30 2021 11:19, Alfonso Montero wrote:
> Also cc'ing @slotThe : This makes me wonder if you're aware and
> confortable with `kmonad` linking to `musl` instead of `glibc`. Can't
> this cause hard-to-reproduce errors on developer's build environments?

I am aware and I'm actually also using musl on my personal computer.
The situation is actually not that bad (once you tell stack that you
_really_ want to use your system's GHC version) and I don't think I've
ever run into any serious troubles related to that.

In either case, statically linking Haskell executables against musl is
quite well established in the community at this point; I'm sure it won't
cause us any further troubles

david-janssen:
Actually, I think that is exactly how we used to build static executables. It was using a nix-script and `static-haskell-nix`, and that, internally, IIRC, was using musl. Seemed to work fine.

@slotThe Does this mean that using docker to build `kmonad` now automatically generates a static executable? This will become relevant once we want to do another release, since I haven't statically built haskell in ages :-).

slotThe:
> @slotThe Does this mean that using docker to build `kmonad` now automatically generates a static executable? This will become relevant once we want to do another release, since I haven't statically built kmonad in ages :-).

Indeed, AFAIK this is the case now!

PR #343, opened 2021-08-21
Add GitHub Actions CI

slotThe:
This adds rudimentary CI support to the project, with the help of [haskell-ci](https://github.com/haskell-CI/haskell-ci).  Right now, it covers GHC 8.6.5, 8.8.4, and 8.10.4.  These are triggered on pull requests and pushes. 

There is also a commit containing a `cabal format`, because the CI did not like certain decisions that were made (haskell/cabal#7018).  Besides, using an autoformatter for this kind of thing seems sensible (to me).

A future direction we could take this in would be to 

  1. compile the CI with `-Werror`,
  2. automatically install and run `hlint`.

Both of these, however, first require us to clean up the current warnings (which is certainly doable, though beyond the imagined scope of this PR).


lierdakil:
Oh boy. Sorry, are you aware of https://github.com/haskell/actions/tree/main/setup#readme ? What I see here looks way too overcomplicated.

slotThe:
> Oh boy. Sorry, are you aware of https://github.com/haskell/actions/tree/main/setup#readme ? What I see here looks way too overcomplicated.

Clearly, I was not :) I went with `haskell-ci` because auto-generating this kind of thing seemed like the most robust way of doing things.  Then again, it probably also is the most annoying to debug when something breaks.  

We may well just copy what xmonad does with regards to a [stack-based solution](https://github.com/xmonad/xmonad/blob/master/.github/workflows/stack.yml) and call it a day

lierdakil:
> We may well just copy what xmonad does with regards to a stack-based solution and call it a day

That sounds reasonable, but I would argue for using Haskell setup action instead of manually setting up the ppa. At least that works not only on Linux, but other platforms as well. I can throw together a workflow script and make a PR if you like.

On a related note, do we really need to test with multiple resolvers/GHC versions?

slotThe:
> That sounds reasonable, but I would argue for using Haskell setup action instead of manually setting up the ppa.

If we immediately get a CI setup on other platforms due to that then I'm all for it.  If not then I don't really have a strong opinion either way; both seem quite fine to me.

> I can throw together a workflow script and make a PR if you like.

Sure, that can supersede this one then.

> On a related note, do we really need to test with multiple resolvers/GHC versions?

I do think it's quite sensible to be backwards compatible with a few version of GHC.  This keeps PRs from accidentally introducing e.g. language extensions that are only available in the newest release of GHC, which not all distributions may package yet.

lierdakil:
> I do think it's quite sensible to be backwards compatible with a few version of GHC

But if we're using stack, it pins and auto-installs the appropriate ghc version anyway. So unless we really want to support multiple stack resolvers for whatever reason, ghc versions are not our concern really.

lierdakil:
> If we immediately get a CI setup on other platforms

Possibly on Windows. Since macOS needs additional dependencies (kext, dext), I suspect it won't be that easy.

slotThe:
> But if we're using stack, it pins and auto-installs the appropriate ghc version anyway (unless specifically instructed not to). So unless we really want to support multiple stack resolvers for whatever reason, ghc versions are not our concern really.

I know of many people who like to use the GHC version that their system provides, opting to set `system-ghc: true` in their stack config file and then adjusting the resolver when necessary.  I think it makes sense for us to try and support multiple resolvers for that reason; it's a small price to pay for something that could add a lot of convenience for users

MuhammedZakir:
This is how Carp does -- https://github.com/MuhammedZakir/Carp/blob/master/.github/workflows/carp.yml. We only need to change/remove the testing commands to use it.

lierdakil:
> I know of many people who like to use the GHC version that their system provides

Eh... I won't really argue this point, but just my 2¢. In my admittedly limited experience, most people familiar with Haskell on Linux use ghcup, and macOS users tend to use homebrew. Both are pretty good at providing recent GHC releases. And on Windows, chocolatey, scoop, or indeed stack is usually used to manage GHC versions -- the former two tend to provide the latest available by default.

People not familiar with Haskell on the other hand don't generally know enough about stack's idiosyncrasies to coerce it to use system GHC and not blow up.

Supporting multiple stack resolvers/GHC versions does come with a pretty heavy cost, by the way, due to API changing over time, both for GHC and dependencies. You get warnings/errors on certain resolvers that when fixed break things on other resolvers, so the code eventually becomes a horrible CPP mess. The cost is justifiable for libraries, but I don't think the same reasoning can be applied to binaries. I tried with pandoc-crossref and at some point had to give up. Won't stop you from trying, though.

slotThe:
On Sat, Aug 21 2021 09:23, Nikolay Yakimov wrote:
> Eh... I won't really argue this point, but just my 2¢. In my
> admittedly limited experience, most people familiar with Haskell on
> Linux use ghcup, and macOS users tend to use homebrew. Both are pretty
> good at providing recent GHC releases. And on Windows, chocolatey,
> scoop, or indeed stack is usually used to manage GHC versions -- the
> former two tend to provide the latest available by default.
>
> People not familiar with Haskell on the other hand don't generally
> know enough about stack's idiosyncrasies to coerce it to use system
> GHC and not blow up.

Just to give a specific example; the situation when using musl libc on a
distro that's not Alpine is still somewhat... spotty.  The easiest
solution (at least that I know) is to simply use the system ghc and
override the resolver when necessary. 

> Supporting multiple stack resolvers/GHC versions does come with a
> pretty heavy cost, by the way, due to API changing over time, both for
> GHC and dependencies. You get warnings/errors on certain resolvers
> that when fixed break things on other resolvers, so the code
> eventually becomes a horrible CPP mess. The cost is justifiable for
> libraries, but I don't think the same reasoning can be applied to
> binaries. I tried with pandoc-crossref and at some point had to give
> up.

In xmonad we (so far) manage quite well with just disabling certain
warnings in the cabal file [1].  I agree that sprinkling CPP throughout
the code would be a step backwards and at that point it's probably not
worth it anymore to support multiple GHC versions.  However, I'd say
we'll cross that bridge when we get to it.

> Won't stop you from trying, though.

I'm fully prepared to eat my own words here, but I at least want to give
this a try before giving up :)

[1]: https://github.com/xmonad/xmonad/blob/master/xmonad.cabal#L87-L88

slotThe:
Closed via #345

PR #345, opened 2021-08-21
Add Haskel CI workflow

lierdakil:
This seems to work. Although I've just realized there aren't any automated tests, so "build and test" step is actually just "build".

macOS needs some additional arguments during build, I've opted to use a separate step instead of bash-fuing it. This leads to some duplication, but it's less obscure, I think.

Build fails on the latest nightly resolver due to changes in RIO API. I've allowed failures on nightly for that reason (whether GitHub UI picks up on this is anyone's guess -- it does on the [action page on my fork](https://github.com/lierdakil/kmonad/actions/runs/1154101823), but not here apparently)

I'm skipping local msys install on windows, which saves a minute or so in build time, and everything builds apparently, but ghc does complain about missing paths a fair bit.

One caveat is that changing/adding resolvers requires manually adding a corresponding GHC version to a JSON string on line 26. Could just let stack install the ghc version it wants, but it takes a while. Didn't think of a better way to handle this, suggestions are welcome. As to why not do it like in xmonad, that's because I don't want to define all (3 OSes)×(3 resolvers) variants (potentially more if you want to add more resolvers) manually, and we'd have to with `include`.
**EDIT:** actually, we can get GHC version from stackage. If stackage is down though, our CI is too.

**EDIT2:** Apparently stackage lts-18.7 just landed which uses GHC 8.10.6. Not yet available for Windows, so I've pinned minor lts versions, which seems like a good idea in general -- don't really want CI to suddenly start failing when nothing has changed on our side.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -81,7 +81,7 @@ jobs:
             --fast --no-terminal --skip-msys \
             --resolver=${{ matrix.resolver }} --system-ghc
 
-      - name: Build and test (macOS)
+      - name: Build and test (macOS, dext)
         shell: bash
         if: runner.os == 'macOS'
         run: |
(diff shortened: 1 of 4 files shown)


slotThe:
This looks great, thanks!

> Build fails on the latest nightly resolver due to changes in RIO API. I've allowed failures on nightly for that reason (whether GitHub UI picks up on this is anyone's guess -- it does on the [action page on my fork](https://github.com/lierdakil/kmonad/actions/runs/1154101823), but not here apparently)

[This patch](http://ix.io/3wDH) should fix the issues on GHC 9.  I also think we should include GHC 8.6.5 in the builds, as it's not that old (indeed, afaik debian is still on 8.4.x, but I think we're already incompatible with that version). 

> **EDIT:** actually, we can get GHC version from stackage. If stackage is down though, our CI is too.

Seems like a decent compromise to me

lierdakil:
> This patch should fix the issues on GHC 9

Seems to do the trick

> include GHC 8.6.5

The common wisdom dictates three last releases, but sure.

slotThe:
> The common wisdom dictates three last releases, but sure.

Perhaps, but I don't think you can really call GHC 9 "released" yet when the whole ecosystem is still on (or even just moving to) 8.10

Thank you for taking care of this!

Issue #330, opened 2021-08-11
Looking For: Additional MacOS Maintainers


slotThe:
Been using kmonad on MacOS?  Want to contribute to the project but don't know how?  Then we want _you_ to... maintain everything for us :)

Jokes aside, @thoelze1—who has been solely responsible for MacOS maintenance up until now—has become quite busy in these last few weeks and so we are looking for additional MacOS maintainers to help out.  This does involve writing a bit of Haskell code, but a lot of the work has also been done in C/C++. 

The first goal would be to actually get #259 (which is a vast simplification of keycode handling) done in order to move forward with improving other parts of kmonad and perhaps even prepare for a new release.  I will look over the Linux parts and David said he'd try to find time for the Windows side of things soon, but neither of us owns a Mac.  

If push comes to shove, there is always the option of just merging #259 with broken MacOS support and waiting for someone to come along and get annoyed enough to fix this, but this is a situation I imagine everyone would like to avoid.  So if you, dear reader, feel up to the task, feel free to contact me or David here or directly.

MuhammedZakir:
I only have a very little Haskell and C++ experience. So I won't even say "I will give it a try" as I am not sure I would be able to do it. But still, what do I need to do? Specifically, what has changed in `keycode-refactor` branch? Any, and all details would make my work a little bit easier. :-)

P.S. If anyone else is interseted, feel free to take this.

Vermoot:
I'd love to be of any help, but every time I look at anything Haskell I feel like I'm having a seizure. I might try to learn some basic Haskell to be able to at least understand KMonad's code, but it'll be a while before I can do anything productive, I fear :/

molleweide:
I am in the same boat as Vermoot. I have been studying haskell during the summer but I am nowhere near
a state of being productive with haskell.

slotThe:
> I only have a very little Haskell and C++ experience. So I won't even say "I will give it a try" as I am not sure I would be able to do it. But still, what do I need to do? Specifically, what has changed in `keycode-refactor` branch? Any, and all details would make my work a little bit easier. :-)
> 
> P.S. If anyone else is interseted, feel free to take this.

Well, the tl;dr is "look what the PR did with regards to [these modules](https://github.com/kmonad/kmonad/pull/259/files#diff-521ee437c2df35fe0cfccfad53bd038644d038b4c79401adfb996f56f9e96cc5L99-R144) and do the same thing for MacOS" :)

[...]

PR #259, opened 2021-05-18
Keycode refactor

david-janssen:
So, I've got the first working version of KMonad with the new Keycode model running on Linux.

The code is still an undocumented mess, and there are warnings that I have to take care of, and then there is more stuff to be done. *BUT* It compiles, it runs, it works.

Next obvious steps:
- Get Windows keycodes implemented (Will take care of this myself)
- Get Mac keycodes implemented (I was hoping @thoelze1 would have a go)
- Get all the aliases added back in so that we don't break backwards support

Once that is done, I think we merge into master.

A lot has changed, I am happy discussing things in this Pull Request. If someone would like to hop into Discord or IRC, then let's arrange a time, and I will show up.

I'll get started on ironing out the warts.

PR #346, opened 2021-08-23
Mac keycode refactor

MuhammedZakir:
Ref: https://github.com/kmonad/kmonad/pull/259
Resolves: https://github.com/kmonad/kmonad/issues/330

### DIFF
--- a/src/KMonad/App/KeyIO/Common.hs
+++ b/src/KMonad/App/KeyIO/Common.hs
@@ -17,7 +17,7 @@ import KMonad.Util.Name
 -- The configuration records for all of the Linux KeyIO options
 
 -- | The configuration record for evdev key-input on Linux
-data EvdevCfg = EvdevCfg
+newtype EvdevCfg = EvdevCfg
   { _evdevPath :: FilePath -- ^ The path to the input-file to open and capture
   } deriving Show
 makeClassy ''EvdevCfg
--- a/src/KMonad/App/KeyIO/Common.hs
+++ b/src/KMonad/App/KeyIO/Common.hs
@@ -50,13 +50,18 @@ instance Default UinputCfg where
 --------------------------------------------------------------------------------
 -- $cfgs-mac
 
--- TODO: This is where the Mac config records go
+-- | Holds the product string of keyboard if given. Else, Nothing.
+newtype IOKitCfg = IOKitCfg
+  { _productStr :: Maybe Text
+  } deriving Show
+makeClassy ''IOKitCfg
 
--- | Placeholder
-data KIOKitCfg = KIOKitCfg deriving Show
+-- | By default, no product string (will grab all keyboards).
+instance Default IOKitCfg where
+  def = IOKitCfg Nothing
 
--- | Placeholder
-data KextCfg = KextCfg deriving Show
+-- | No config options for Kext/Dext
+data ExtCfg = ExtCfg deriving Show
 
 --------------------------------------------------------------------------------
 -- $cfgs-win
@@ -76,14 +81,14 @@ data SendEventCfg = SendEventCfg deriving Show
 
 data KeyInputCfg
   = LinuxEvdevCfg          EvdevCfg
-  | MacKIOKitCfg           KIOKitCfg
+  | MacIOKitCfg            IOKitCfg
   | WindowsLowLevelHookCfg LowLevelHookCfg
   deriving Show
 
 
 data KeyOutputCfg
   = LinuxUinputCfg      UinputCfg
-  | MacKextCfg          KextCfg
+  | MacExtCfg           ExtCfg
   | WindowsSendEventCfg SendEventCfg
   deriving Show
 
--- a/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
+++ b/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
@@ -45,11 +45,10 @@ foreign import ccall "wait_key"
 waitKey :: Ptr RawEvent -> OnlyIO RawEvent
 waitKey ptr = do
        re <- wait_key ptr >> peek ptr
-       -- Filter `Keyboard_Reserved` keycode sent on each key event.
-       -- E.g. tap a -> P(0x7,0xFFFFFFFF) Pa Ra R(0x7,0xFFFFFFFF)
+       -- Filter unwanted keycodes sent on each key event.
        case (re^.reCode) of
          (0x7, 0xFFFFFFFF) -> waitKey ptr
-         -- (0x7, 0x1)        -> waitKey ptr
+         (0x7, 0x1)        -> waitKey ptr
          _                 -> return re
 
 -------------------------------------------------------------------------------
--- a/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
+++ b/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
@@ -45,11 +45,12 @@ foreign import ccall "wait_key"
 waitKey :: Ptr RawEvent -> OnlyIO RawEvent
 waitKey ptr = do
        re <- wait_key ptr >> peek ptr
-       -- Filter `Keyboard_Reserved` keycode sent (only?) by
-       -- Apple keyboards on each key event.
-       -- E.g. tap a -> P(0x7,0xFFFF) Pa Ra R(0x7,0xFFFF)
-       if (0x7, 0xFFFF) == (re^.reCode) then waitKey ptr
-                                        else return re
+       -- Filter `Keyboard_Reserved` keycode sent on each key event.
+       -- E.g. tap a -> P(0x7,0xFFFFFFFF) Pa Ra R(0x7,0xFFFFFFFF)
+       case (re^.reCode) of
+         (0x7, 0xFFFFFFFF) -> waitKey ptr
+         -- (0x7, 0x1)        -> waitKey ptr
+         _                 -> return re
 
 -------------------------------------------------------------------------------
 
@@ -82,7 +83,7 @@ withIOKitSource cfg = mkCtx $ \f -> do
   let nextEvent :: IO m => Ptr RawEvent -> m KeySwitch
       nextEvent ptr = do
         re <- liftIO $ waitKey ptr
-        pure $ mkKeySwitch (if re^.reVal == 0 then Press else Release)
+        pure $ mkKeySwitch (if re^.reVal == 0 then Release else Press)
                            (Keycode $ re^.reCode)
 
   bracket init cleanup $ \ptr -> f (nextEvent ptr)
--- a/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
+++ b/src/KMonad/App/KeyIO/Mac/IOKitSource.hs
@@ -1,0 +1,88 @@
+{-|
+Module: KMonad.App.KeyIO.Mac.IOKitSource
+Description: Use mac c-api to grab keyboard(s).
+TODO: complete this
+-}
+
+module KMonad.App.KeyIO.Mac.IOKitSource
+  ( withIOKitSource )
+where
+
+import KMonad.Prelude
+
+import Foreign.C.String
+import Foreign.Ptr
+import Foreign.Marshal hiding (void)
+import Foreign.Storable
+
+import KMonad.App.KeyIO.Common
+import KMonad.Util
+import KMonad.Util.Keyboard.Mac
+
+
+-------------------------------------------------------------------------------
+-- $ffi
+
+-- | Use the mac c-api to grab keyboard(s)
+foreign import ccall "grab_kb"
+  grab_kb :: CString -> OnlyIO Word8
+
+grabKb :: CString -> OnlyIO ()
+grabKb = void . grab_kb
+
+-- | Release the keyboard hook
+foreign import ccall "release_kb"
+  release_kb :: OnlyIO Word8
+
+releaseKb :: OnlyIO ()
+releaseKb = void release_kb
+
+-- | Pass a pointer to a buffer to wait_key, and when it
+-- returns, the buffer can be read for the next key event.
+foreign import ccall "wait_key"
+  wait_key :: Ptr RawEvent -> OnlyIO Word8
+
+waitKey :: Ptr RawEvent -> OnlyIO RawEvent
+waitKey ptr = do
+       re <- wait_key ptr >> peek ptr
+       -- Filter `Keyboard_Reserved` keycode sent (only?) by
+       -- Apple keyboards on each key event.
+       -- E.g. tap a -> P(0x7,0xFFFF) Pa Ra R(0x7,0xFFFF)
+       if (0x7, 0xFFFF) == (re^.reCode) then waitKey ptr
+                                        else return re
+
+-------------------------------------------------------------------------------
+
+withIOKitSource :: LUIO m e => IOKitCfg -> Ctx r m GetKey
+withIOKitSource cfg = mkCtx $ \f -> do
+
+  let prodStr :: Maybe Text
+      prodStr = cfg^.productStr
+
+  let init :: LUIO m e => m (Ptr RawEvent)
+      init = do
+        case prodStr of
+          Nothing -> do
+            logInfo "Opening HID devices"
+            liftIO $ grabKb nullPtr
+          Just s -> do
+            logInfo $ "Opening HID device: " <> s
+            liftIO $ withCString (unpack s) grabKb
+
+        liftIO . mallocBytes $ sizeOf (undefined :: RawEvent)
+
+  let cleanup :: LUIO m e => Ptr RawEvent -> m ()
+      cleanup ptr = do
+        case prodStr of
+          Nothing -> logInfo "Closing HID devices"
+          Just s -> logInfo $ "Closing HID device: " <> s
+
+        liftIO $ releaseKb >> free ptr
+
+  let nextEvent :: IO m => Ptr RawEvent -> m KeySwitch
+      nextEvent ptr = do
+        re <- liftIO $ waitKey ptr
+        pure $ mkKeySwitch (if re^.reVal == 0 then Press else Release)
+                           (Keycode $ re^.reCode)
+
+  bracket init cleanup $ \ptr -> f (nextEvent ptr)
--- a/src/KMonad/Util/Keyboard/Mac.hs
+++ b/src/KMonad/Util/Keyboard/Mac.hs
@@ -1,0 +1,197 @@
+module KMonad.Util.Keyboard.Mac
+  ( -- * Keycode
+    Keycode(..)
+
+    -- * Event types
+  , EvType(..)
+  , _EvType
+
+    -- * Raw event as defined in Mac
+  , RawEvent(..)
+  , HasRawEvent(..)
+  , mkRaw
+
+    -- * Names for keycodes
+  , keycodeNames
+  )
+
+where
+
+
+import KMonad.Prelude
+
+import Numeric (showHex)
+import Data.Char (toUpper)
+
+import Foreign.Storable
+import KMonad.Util.Keyboard.Common
+import KMonad.Util.Keyboard.Keynames
+
+import qualified RIO.HashMap as M
+import qualified RIO.Text    as T
+import RIO.List.Partial ((!!))
+
+
+--------------------------------------------------------------------------------
+-- $keycode
+
+-- first Word32 represents the IOKit usage page
+-- second Word32 represent the IOKit usage
+newtype Keycode = Keycode { unKeycode :: (Word32, Word32) }
+  deriving (Eq, Ord, Hashable)
+
+instance Show Keycode where
+  show (Keycode (p, u)) = "(" ++ toHex p ++ ", " ++ toHex u ++ ")"
+
+toHex :: Word32 -> String
+toHex i = "0x" ++ (map toUpper $ showHex i "")
+
+--------------------------------------------------------------------------------
+-- $evtype
+
+-- | Mac has two types types of events: Press and Release.
+--
+-- The type is ordered in a way to line up with the integers our Mac API uses to
+-- encode for these events. I.e. 0=Press, 1=Release
+data EvType = MacPress | MacRelease
+  deriving (Eq, Show, Enum)
+
+-- | Prism between Mac and Haskell representation of event types
+_EvType :: Prism' Word64 EvType
+_EvType = prism' (fi . fromEnum) $ \i ->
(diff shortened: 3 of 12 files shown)


MuhammedZakir:
For this to work, comment out `cmps` and `102d` keycode aliases in `src/KMonad/App/Parser/Keycode.hs`. Non-common aliases should be moved to OS-specific files?

src/KMonad/App/KeyIO/Mac/IOKitSource.hs
```diff
@@ -0,0 +1,72 @@
...
+        case (prodStr) of
+          Nothing -> do
+            logInfo "Opening HID devices"
+            liftIO . void $ grab_kb nullPtr
+          Just s -> do
+            logInfo $ "Opening HID device: " <> s
+            liftIO $ do
+              str <- newCString $ unpack s
+              void $ grab_kb str
+              free str
```
slotThe, line 56: Everytime you have this `newCString` -> do stuff -> `free` pattern going on its worth looking at whether `withCString` may be a better fit

src/KMonad/App/KeyIO/Mac/Kext.hs
```diff
@@ -0,0 +1,43 @@
...
+  send_key :: Ptr RawEvent -> OnlyIO ()
+
+-------------------------------------------------------------------------------
+
+withKext :: LUIO m e => KextCfg -> Ctx r m PutKey
+withKext _ = mkCtx $ \f -> do
+
+  let init = do
+        logInfo "Initializing Mac key sink"
+        liftIO $ mallocBytes $ sizeOf (undefined :: RawEvent)
```
slotThe, line 30: Same comment applies here with regards to `undefined` vs. `Proxy`

src/KMonad/Util/Keyboard/Mac.hs
```diff
@@ -0,0 +1,197 @@
...
+
+-- https://stackoverflow.com/questions/1959715/how-to-print-integer-literals-in-binary-or-hex-in-haskell/55623656#55623656
+toHex :: Word32 -> String
+toHex = ("0x" ++) . convertToBase 16
+  where
+    convertToBase :: Word8 -> Word32 -> String
+    convertToBase b n
+      | n < 0              = '-' : convertToBase b (-n)
+      | n < fromIntegral b = [(['0'..'9'] ++ ['A' .. 'Z']) !! fromIntegral n]
+      | otherwise          = let (d, m) = n `divMod` fromIntegral b in convertToBase b d ++ convertToBase b m
```
slotThe, line 51: What's the reasoning for prefering this to `Numeric.showHex`?

src/KMonad/Util/Keyboard/Mac.hs
```diff
@@ -0,0 +1,197 @@
...
+mkRaw p c = RawEvent
+  { _reCode = unKeycode c
+  , _reVal  = p ^. re _EvType }
+
+
+--------------------------------------------------------------------------------
+-- $keynames
+
+-- | zip 2 lists only if they have the same length.
+--   Caution: only ever use this to generate static data.
```
slotThe, line 109: I think this is a bit misleading, as it suggests that the only thing that's happening is a length comparison and then a `zip`

src/KMonad/Util/Keyboard/Mac.hs
```diff
@@ -0,0 +1,197 @@
...
+z :: Word32 -> [Word32] -> [Text] -> [(CoreName, Keycode)]
+z page codes names
+  | length codes == length names = zipWith func codes names
+  | otherwise            = error "mac: keycodes length"
+  where
+    func = (\c n -> (CoreName n, Keycode (page, c)))
```
slotThe, line 115:
```suggestion
    func c n = (CoreName n, Keycode (page, c))
```

src/KMonad/Util/Keyboard/Mac.hs
```diff
@@ -0,0 +1,197 @@
...
+  , z 0xFF01 [0x20..0x21] [ "Vbru", "Vbrd" ]
+
+  -- ###  0xFF => AppleVendor Page Top Case  ###
+  , z 0xFF [0x3..0x5] [ "Fn", "brup", "brdn" ]
+  -- kbil -> IlluminationToggle
+  , z 0xFF [0x7..0x9] [ "kbil", "kbup", "kbdn" ]
+  ]
+   where
+      -- Apple uses "1-9 & 0" order for numbers
+      num_1To9 = map tshow $ [1..9 :: Int]
```
slotThe, line 184:
```suggestion
      num_1To9 = map tshow [1..9 :: Int]
```

src/KMonad/Util/Keyboard/Mac.hs
```diff
@@ -0,0 +1,197 @@
...
+toHex :: Word32 -> String
+toHex = ("0x" ++) . convertToBase 16
+  where
+    convertToBase :: Word8 -> Word32 -> String
+    convertToBase b n
+      | n < 0              = '-' : convertToBase b (-n)
+      | n < fromIntegral b = [(['0'..'9'] ++ ['A' .. 'Z']) !! fromIntegral n]
+      | otherwise          = let (d, m) = n `divMod` fromIntegral b in convertToBase b d ++ convertToBase b m
+--------------------------------------------------------------------------------
+  -- $evtype
```
slotThe, line 53:
```suggestion
-- $evtype
```

Otherwise Haddock is confused (at least on my system)

slotThe (commented):
Thank you for taking the initiative!  Due to not having a Mac, I can only review the code as-is (that is, without testing it).  

Cc. @david-janssen for a possible inclusion in #259

src/KMonad/App/KeyIO/Common.hs
```diff
@@ -50,12 +50,17 @@ instance Default UinputCfg where
 --------------------------------------------------------------------------------
 -- $cfgs-mac
 
--- TODO: This is where the Mac config records go
+-- | Holds the product string of keyboard if given. Else, Nothing.
+data IOKitCfg = IOKitCfg
```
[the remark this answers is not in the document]
MuhammedZakir: Should I also change `EvdevCfg`?

src/KMonad/Util/Keyboard/Mac.hs
MuhammedZakir: What would be a better description?

src/KMonad/App/KeyIO/Common.hs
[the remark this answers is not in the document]
slotThe: I would say so, yes

src/KMonad/Util/Keyboard/Mac.hs
slotThe: Good question :)  Since this is not polymorphic, I guess we could write exactly what's going on; something about building up named keycode pairs, perhaps

MuhammedZakir:
Two questions:

1.  > Apple internal keyboard wraps key events with (iirc) `(0x7, 0xFF)` keycode. For example, when tapping `a`, instead of `Pa Ra`, we get `P(0x7, 0xFF) Pa Ra R(0x7, 0xFF)`. IIAR, we must ignore this keycode when capturing events, but I am an not sure whether we should emit this when sending events. FWIW, I used Karabiner-EventViewer for testing, and I don't get this when I use Karabiner-Elements.
    > 
    > tl;dr -- should we wrap key [send] events with key event of this keycode, or should we ignore it?

2. Although we support both Kext and Dext and the config for both is the same, the word "Kext" is used in many of the names of types and functions. I think it is better to use a general name for clarity, but I can't think of a name. Any suggestion?

src/KMonad/App/KeyIO/Mac/IOKitSource.hs
```diff
@@ -0,0 +1,72 @@
...
+
+        liftIO . mallocBytes $ sizeOf (undefined :: RawEvent)
```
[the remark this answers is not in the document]
MuhammedZakir: Didn't know about that. Thanks! :-)

src/KMonad/Util/Keyboard/Mac.hs
MuhammedZakir:
```
-- | Takes a usage page, a list of usages and a list of names, 
--    and returns a list of tuples of keycodes and their names.
--
-- Caution: Only use this for static data as this throws an error
-- if the lengths of lists differ.
```

I am not familiar with Haddock; if there is any mistake in markup, feel free to correct me.

MuhammedZakir:
In this amended commit, I have also changed `waitKey` to filter `(0x7, 0xFFFF)` keycode. I coverted this PR to draft as I have to do a few tests regarding this. Will change it back once that's done.

src/KMonad/Util/Keyboard/Mac.hs
slotThe: Looks good to me :+1:

slotThe:
Sorry for not getting back to you sooner; I've been having lots of IRL things coming up with trying to finish my thesis. 

>     1. > Apple internal keyboard wraps key events with (iirc) `(0x7, 0xFF)` keycode. For example, when tapping `a`, instead of `Pa Ra`, we get `P(0x7, 0xFF) Pa Ra R(0x7, 0xFF)`. IIAR, we must ignore this keycode when capturing events, but I am an not sure whether we should emit this when sending events. FWIW, I used Karabiner-EventViewer for testing, and I don't get this when I use Karabiner-Elements.
>        > tl;dr -- should we wrap key [send] events with key event of this keycode, or should we ignore it?

Karabiner is the de facto standard remapping utility for Mac, right?  In this case I think emulating what they do can't be _that_ bad

>     2. Although we support both Kext and Dext and the config for both is the same, the word "Kext" is used in many of the names of types and functions. I think it is better to use a general name for clarity, but I can't think of a name. Any suggestion?

Are these shorthand for kernel extensions and driver extensions?  In this case something like like just extensions might fit the bill (but I'm bad with names so don't listen to me :))

MuhammedZakir:
> Sorry for not getting back to you sooner; I've been having lots of IRL things coming up with trying to finish my thesis.

Np. Good luck with your thesis! :-)

> Karabiner is the de facto standard remapping utility for Mac, right? In this case I think emulating what they do can't be _that_ bad

So we will ignore that keycode.

> Are these shorthand for kernel extensions and driver extensions? In this case something like like just extensions might fit the bill (but I'm bad with names so don't listen to me :))

I will use `Ext`. Can change if someone gives a better name.

MuhammedZakir:
I got some compilation errors after the latest change. I probably won't be able to fix them in coming days. I will post here after they are fixed.

MuhammedZakir:
I think I have fixed the compilation errors except when I use `sizeOf (Proxy @RawEvent)` instead of `sizeOf (undefined :: RawEvent), I get

```
    • Could not deduce (Storable (Proxy RawEvent))
        arising from a use of ‘sizeOf’
      from the context: LUIO m e
        bound by the type signature for:
                   withIOKitSource :: forall (m :: * -> *) e r.
                                      LUIO m e =>
                                      IOKitCfg -> Ctx r m GetKey
        at /home/user/dev/haskell/kmonad/src/KMonad/App/KeyIO/Mac/IOKitSource.hs:59:1-57
    • In the second argument of ‘($)’, namely
        ‘sizeOf (Proxy @RawEvent)’
      In a stmt of a 'do' block:
        liftIO . mallocBytes $ sizeOf (Proxy @RawEvent)
      In the expression:
        do case prodStr of
             Nothing -> do ...
             Just s -> do ...
           liftIO . mallocBytes $ sizeOf (Proxy @RawEvent)
   |
74 |         liftIO . mallocBytes $ sizeOf (Proxy @RawEvent)
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^
```

slotThe:
> I think I have fixed the compilation errors except when I use `sizeOf (Proxy @RawEvent)` instead of `sizeOf (undefined :: RawEvent)`, I get
> [...]

Whoops, sorry about that; I must have misread the type signature of `sizeOf`.  In this case, `Proxy` will not work and `undefined` is the way to go

MuhammedZakir:
> Whoops, sorry about that; I must have misread the type signature of `sizeOf`. In this case, `Proxy` will not work and `undefined` is the way to go

Np.

---

Now the only thing left to do is to do some tests. It will take a few days. I will post here once that's done.

nightscape:
I'm just testing this on [my BONE keyboard layout](https://github.com/nightscape/kmonad/blob/nightscape/keymap/user/nightscape/bone_mac.kbd) and it's throwing the following error:
```
kmonad: NonExistentTarget ("compose","cmps")
```
Surprising thing is that I didn't find anything in the diffs of this PR regarding `compose` or `cmps`...
@MuhammedZakir did you ever encounter this problem?

MuhammedZakir:
@nightscape: It is because of aliases.

https://github.com/kmonad/kmonad/pull/346#issuecomment-903878980:
> For this to work, comment out keycode aliases for `cmps` and `102d` in `src/KMonad/App/Parser/Keycode.hs`. 

---

@slotThe @david-janssen :
> Aliases for key codes that are not in all 3 OSs should be moved to OS-specific files?

david-janssen:
Hey there,

Sorry for my loooooong absenses, both in the past and probably in the future. Thank you very much for taking point on this, I wouldn't know where to start with Mac stuff.

> Edit: Simples solution is to move OS-specific aliases in Util/Keyboard/<os>.hs, and then merge them into keycodeAliases in App/Parser/Keycode.hs. I guess this could work. What do you think?

That seems like a good solution. It's been a few months since I dove into code, so it's all a bit hazy at the moment. Trying to get my bearings at the moment. IIRC, the plan was to have a single set of names for keys to use internally, to provide a collection of aliases inside the parser, and document very carefully which names should work cross-platform and which shouldn't.

The aliases should be set up in such a way that a standard 100% keyboard (ISO or ANSI) can be easily described for any OS.

@MuhammedZakir If there is anywhere you'd like me to look at first, let me know. I'm hoping to start dripping a little bit of time into KMonad so long as my health holds.

david-janssen:
Huzzah! I managed (after great headaches) to actually get a MacOS virtualbox up and running, and even to install the `dext` and compile KMonad. I can run KMonad, and looking at the verbose logging, it correctly captures keyboard input and internally processes it well enough, but I can't get KMonad to produce keyboard output.

I have no idea if this is because I'm running inside a virtualbox, or if there is some other reason, and I wouldn't know where to begin with debugging this issue.

In regards to the aliases, the 2 broken ones are:
'compose' and '102d'. We do need to figure out how you call those in OSX, because they refer to buttons that exist on a standard 100% keyboard. 102nd is the key between `lshift` and `z` on ISO-keyboards. `compose` is the key between `alt gr` or `right-alt` and `rctrl`. We need to figure out some naming-scheme that is supported across all 3 platforms to address every single key on 100% keyboards.

I'll move on to working on the windows keycode refactor for now and wait for you to respond.

Cheerio

MuhammedZakir:
Have you allowed `Accessibility` for the terminal/app you are using to run KMonad? Is your installed `Dext`'s version correct? Also, I don't remember, but I think you may have to run KMonad using sudo.

---

I don't have an ISO keyboard, so this is just my guess: I think the key between `lshift` and `z` is `(0x7, 0x64) - NonUS Backslash` and the `compose` key is probably `(0x7, 0x65) - Application`. If you (or someone else) have ISO keyboard, then you can get the keycode using Karabiner-EventViewer.

david-janssen:
> I don't have an ISO keyboard, so this is just my guess: I think the key between lshift and z is (0x7, 0x64) - NonUS Backslash and the compose key is probably (0x7, 0x65) - Application. If you (or someone else) have ISO keyboard, then you can get the correct keycode using Karabiner-EventViewer.

Those both sound very plausible. I don't have either an ISO keyboard I can plug into anything, but maybe I can ask around on the Karabiner github or something.

> Have you allowed Accessibility for the terminal/app you are using to run KMonad? Is your installed Dext's version correct? Also, I don't remember, but I think you may have to run KMonad using sudo.

I needed to go through all of those steps just to get KMonad to run, and I followed the Mac installation instructions to the letter. I think, however, I'm basically just happy that I can see it compiles correctly without intending to tinker with Mac alot myself. It could very well be an issue with running in a virtualbox, since Mac already makes that very annoying, and I think, technically in breach of the user-agreement...

How should we proceed from here? Shall I have a go at merging this PR in or are there any remaining issues that you are aware of on the Mac side of things? After we merge this in I have some serious tinkering to do with the Windows stuff, that code isn't even finished yet, but I have had success in the past developing through VirtualBox.

MuhammedZakir:
> [...] but maybe I can ask around on the Karabiner github or something.

I guess we'll have to do that.

> I needed to go through all of those steps just to get KMonad to run, and I followed the Mac installation instructions to the letter. I think, however, I'm basically just happy that I can see it compiles correctly without intending to tinker with Mac alot myself. It could very well be an issue with running in a virtualbox, since Mac already makes that very annoying, [...]

It was working fine the last time I tried, but I wanted to someone else to try it out, just in case. :-)

About this whole permission issues and stuff: I did get the exact issue you got, but I don't remember what I did to fix it. That's why, I mentioned all three possible fixes I can think of. Anyway, I will undo whatever I have done, and try to "install" KMonad again to verify the installation instructions.

> [...] and I think, technically in breach of the user-agreement...

It is a breach of user-agreement.

---

> How should we proceed from here? Shall I have a go at merging this PR in or are there any remaining issues that you are aware of on the Mac side of things?

There is no "issue" per se. However, I wanted to do one test\* regarding ignoring `Keyboard_Reserved` events. I will do that tomorrow. Other than that, everything is good.

\* Although technically there shouldn't be any problem, I wanted to make sure it's working as intended.

> After we merge this in I have some serious tinkering to do with the Windows stuff, that code isn't even finished yet, but I have had success in the past developing through VirtualBox.

Feel free to ping me if you need any help.

MuhammedZakir:
Don't merge this. I am getting a weird behavior; when tapping `x` for the time, I am getting `P(0x7, 0x1) Rx P(0x7, 0x1) P(0x7, 0x1) Px P(0x7,0x1) Px`. For the subsequent taps of `x`, I am getting `P(0x7, 0x1) Rx P(0x7, 0x1) Rx P(0x7, 0x1) Px P(0x7,0x1) Px`. I am not sure what is the problem here. I am pretty sure I didn't get this the last time I tested. I don't know what went wrong. :-( Need sometime to debug this.

Btw, `(0x7, 0x1)` is `ErrorRollOver`.

david-janssen:
No worries, I have some work to do on the Windows code anyways. I'm very happy leaving OSX stuff in your hands.

MuhammedZakir:
I fixed it. However, I am still getting events for `(0x7, 0x1)`, but it's not being send/"put". That is, if I press `x`, I get
(note: `ErrorRollOver` == `(0x7, 0x1)`)
```
Got: Release ErrorRollOver
Got: Press x
Running Hooks
Running Hooks
Put: Press x
Registering untimed hook: 1
Got: Release ErrorRollOver
Got: Release x
Running hooks
Registering untimed hook: 2
Running hooks
Put: Release x
```

I am confused:
1. Why is `ErrorRollOver` not being send/"put"?
2. Will `ErrorRollOver` release event cause problem for some buttons? Should we ignore it?

---

> > I don't have an ISO keyboard, so this is just my guess: I think the key between lshift and z is (0x7, 0x64) - NonUS Backslash and the compose key is probably (0x7, 0x65) - Application. If you (or someone else) have ISO keyboard, then you can get the correct keycode using Karabiner-EventViewer.
> 
> Those both sound very plausible. I don't have an ISO keyboard I can plug into anything

From https://gist.github.com/MightyPork/6da26e382a7ad91b5496ee55fdc73db2#file-usb_hid_keys-h-L142:
```c
#define KEY_102ND 0x64 // Keyboard Non-US \ and |
#define KEY_COMPOSE 0x65 // Keyboard Application
```

david-janssen:
> 1. Why is `ErrorRollOver` not being send/"put"?
> 2. Will `ErrorRollOver` release event cause problem for some buttons? Should we ignore it?

In regards to `1.`, I'm not entirely sure, since the way that MacOS codes its event-stream is still a mystery to me. However, in regards to `2.`, I think a very firm *yes* here, since `ErrorRollOver` is an event meant to indicate something semantically along the lines of: "Too many keys were pressed at the same time, so I don't really know what is pressed". The only thing we could sanely do with that is simply nothing. 

In the end, the input stream that should arrive at KMonad haskell code is a series of press-release alternations. e.g. `Pa Pb Ra Pc Rb Rc`. I think simply on the merit of the keymap coding we automatically ignore repeated presses or releases, but it might be good to even make it explicit.

> From https://gist.github.com/MightyPork/6da26e382a7ad91b5496ee55fdc73db2#file-usb_hid_keys-h-L142:
> 
> ```c
> #define KEY_102ND 0x64 // Keyboard Non-US \ and |
> #define KEY_COMPOSE 0x65 // Keyboard Application
> ```

Ahhh, great. You confirmed it :-). I'll uncomment the aliases in the `Parser`, did you add the names already to the mac keyboard definition? I suggest we stick with "102d" and "cmps" as the primary internal names for these keys, but we can add aliases like "nonUS" and "app" to the Parser.

david-janssen:
Just to keep you in the loop, I updated all the keycode names for Windows, so now we are compiling fine under windows and a full 100% keyboard works without issue. The last thing I have to fix for windows is key-repeat, but after that both Linux and Windows will be ready. I'm awaiting your final "OK" to merge in the mac stuff, and after that I'll take some time to carefully rebase this ontop of master. Then a bit of testing, and maybe we can roll this out by the end of the week :-)

MuhammedZakir:
In the new commit, `ErrorRollOver` keycode is also filtered. I have also changed keycode names for `non-US \` and `Application` in Mac to `102d` and `cmps`. And, fixed the aliases.

---

> The last thing I have to fix for windows is key-repeat, but after that both Linux and Windows will be ready.

Currently, there in no key-repeat if KMonad is used in Linux virtual console. IIRC, key-repeat is not baked into Windows, and if that's the case, it would be nice if the key-repeat feature is made generic, so that we could use it in Linux, and if need, in Mac too.

---

There is a bug in Mac related to modifier keys when more than 1 KMonad app is run at the same time. Because what needs to be fixed is mainly in C++ side, for now, we can merge this and I will open a new PR when I find a solution for it.

MuhammedZakir:
Btw, do you need me to squash unnecessary commits and rebase them to get a clean working tree?

david-janssen:
> Btw, do you need me to squash unnecessary commits and rebase them to get a clean working tree?

Nah, I was planning on doing that when we merge `keycode-refactor` onto master.

david-janssen:
Wahay!

Issue #326, opened 2021-08-07
The 5 numpad button does not work (alias kp5).


Veanvi:
OS: Windows 10, ver. 20H2

The 5 numpad button does not work (alias kp5), I have to replace it with a button from the num­ber row.
I am not sure that this is the expected Kononad behavior, the remaining numpad buttons work.

MuhammedZakir:
Please share your config.

Veanvi:
I took the settings of Numpad from kmonad templates, but the kp5 key does not work on Windows. Another person who [asked the question](https://github.com/kmonad/kmonad/issues/315) also does not use the five Numpad button in his config, probably he has the same problem on windows.
My config file. 

```
(defcfg
    ;; For Windows
    input  (low-level-hook)
    output (send-event-sink)

    ;; This option tells KMonad to let non-configured keys act normal
    fallthrough true
)

(defsrc
    esc   f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12
    grv   1    2    3    4    5    6    7    8    9    0    -    =    f13
    tab   q    w    e    r    t    y    u    i    o    p    [    ]    ret
    bspc  a    s    d    f    g    h    j    k    l    ;    '    \
    lsft  z    x    c    v    b    n    m    ,    .    /    rsft
    lctl lmet lalt            spc            ralt rmet cmp  rctl
)

(defalias
    ;; Псевдонимы для слоя main ------------------------------------------------

    swq (layer-switch default_qwerty)
    swf (layer-switch main)

    Sf (tap-hold-next-release 200 f lsft)
    Cd (tap-hold-next-release 200 d lctl)
    As (tap-hold-next-release 200 s lalt)
    Ma (tap-hold-next-release 200 a lmet)

    Sj (tap-hold-next-release 200 j rsft)
    Ck (tap-hold-next-release 200 k lctl)
    Al (tap-hold-next-release 200 l lalt)
    M; (tap-hold-next-release 200 ; lmet)

    sfn (tap-hold-next-release 200 spc (layer-toggle spacefn))

    afp (layer-toggle altfn)

    ;; Псевдонимы для слоя spacefn ---------------------------------------------

    Slf (tap-hold-next-release 200 left rsft)
    Cdw (tap-hold-next-release 200 down lctl)
    Art (tap-hold-next-release 200 rght ralt)
    Min (tap-hold-next-release 200 ins lmet)

    f6 (tap-hold-next-release 200 f6 rsft)
[...]

PR #350, opened 2021-08-28
Correcting Windows virtual-key code translation

whitbur:
Kmonad uses a map to associate Windows virtual-key codes with KMonad KeyCodes. When translating the other direction, the logic would flip the entries of this map and go from KeyCode to virtual-key code. However, where more than one virtual-key code mapped to the same KMonad KeyCode, this created a conflict where only one would be able to be sent to the sink, and often not the expected code.

This commit fixes that issue by maintaining two separate maps, one for each direction. In this way, multiple virtual-key codes may be interpreted as the same KMonad KeyCode, and multiple KeyCodes may be sent as the same virtual-key code as necessary.

This is a fix for #326.


src/KMonad/Keyboard/IO/Windows/Types.hs
```diff
@@ -307,6 +314,271 @@ kcMap = M.fromList $
...
   -- , (0xFD, ???)             -- Defined as VK_PA1
-  -- , (0xFE, ???)             -- Defined as VK_CLEAR
+  -- , (0xFE, KeyDelete)       -- Defined as VK_OEM_CLEAR
   ]
 
+-- | Translate a KMonad KeyCode to the corresponding Windows virtual-key code
+-- 
+-- We cannot simply reverse the above map for the opposite direction, because
+-- there will be duplicates where more than one virtual-key code produces the
+-- same KMonad KeyCode. See #326
```
slotThe, line 324: We should probably put a full link to the relevant GitHub issue here instead of hoping every reader knows which forge we use

slotThe (commented):
The code looks fine to me; I'm ready to merge as soon as some people on windows confirm that this works for them.  @Veanvi would you care to test whether this fixes your problem?

Veanvi:
Commit whitbur decided my problem with a non-working the alias kp5. Thanks.

PR #354, opened 2021-09-11
Update Stack to LTS 18.9

nightscape:
GHC 8.10.7 contained in Stack LTS 18.9 allows compiling to native Apple Silicon code.
I've tested this both on an M1 and an Intel chip and both work nicely.
I don't know if there's any downsides to upgrading Stack?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-18.5
+resolver: lts-18.9
 packages:
 - .
 extra-deps:


slotThe:
> I don't know if there's any downsides to upgrading Stack?

Not really, we test with enough stack resolver that we'd notice if something breaks and users are free to adjust this based on their GHC version anyways

Thanks!

PR #361, opened 2021-09-23
Correcting Windows virtual-key code translation

whitbur:
Kmonad uses a map to associate Windows virtual-key codes with KMonad KeyCodes. When translating the other direction, the logic would flip the entries of this map and go from KeyCode to virtual-key code. However, where more than one virtual-key code mapped to the same KMonad KeyCode, this created a conflict where only one would be able to be sent to the sink, and often not the expected code.

This commit fixes that issue by maintaining two separate maps, one for each direction. In this way, multiple virtual-key codes may be interpreted as the same KMonad KeyCode, and multiple KeyCodes may be sent as the same virtual-key code as necessary.

This is a fix for #326.

Previously requested in #350. I switched branches so I had to open a new PR. I've updated based on the previous comments, and at least one other user has confirmed that this fixes the issue.

### DIFF
--- a/src/KMonad/Keyboard/IO/Windows/Types.hs
+++ b/src/KMonad/Keyboard/IO/Windows/Types.hs
@@ -90,12 +90,11 @@ _WinSwitch = iso to' from'
 
 -- | Lookup the corresponding 'Keycode' for this 'WinKeycode'
 fromWinKeycode :: WinKeycode -> Maybe Keycode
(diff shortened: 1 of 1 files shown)


slotThe:
Ah, I must have missed @Veanvi confirming that this works for them.  In that case I don't see a reason to hold out any longer; thanks!

PR #364, opened 2021-09-26
Fix nix build

dlip:
- Flake importing a non-existing file
- Cabal2nix error due to missing name
- Flattened derivation
- Updated nixpkgs


slotThe:
Thanks!

PR #369, opened 2021-10-05
support "+," deadkey sequence for adding a cedilla

gelisam:
The list of supported diacritical marks is hardcoded, so e.g. `#(+' e)`
produces `é` but `#(+, c)` was a syntax error. It now produces `ç`.


gelisam:
By the way, `CONTRIBUTING.md` says to make PRs against the `develop` branch, but I did not see any branch with that name.

david-janssen:
> By the way, CONTRIBUTING.md says to make PRs against the develop branch, but I did not see any branch with that name.

As usual, documentation lags behind reality. :-). I should deal with that at some point...

Thanks for the PR, gonna let the integration tests do their thing and if nothing gets turned up I approve this for merging. Gonna be travelling soon, so if I forget and some other admin sees that the checks have completed, feel free to just merge this right in.

Issue #371, opened 2021-10-05
Key aliases missing after keycode-refactor, breaking user configs


Artefact2:
As of 2356e27, kmonad will not start for me.

Here is the error:
~~~
Oct 05 16:33:36 Silmeria kmonad[1830]: kmonad: Mismatch between length of 'defsrc' and deflayer <default>
Oct 05 16:33:36 Silmeria kmonad[1830]: Source length: 2
Oct 05 16:33:36 Silmeria kmonad[1830]: Layer length: 1
~~~

Here is my config file:
~~~
(defcfg
        input (device-file "/dev/input/by-id/usb-TypeMatrix.com_USB_Keyboard-event-if01")
        output (uinput-sink "kmonad (typematrix 2030 106, media keys)")
        fallthrough true
        allow-cmd false)

(defsrc pp)
(deflayer default lmet)
~~~

slotThe:
On Tue, Oct 05 2021 07:38, Romain D. wrote:
> (defsrc pp)

Is this a real key?  My guess is that it tries to parse this as two `p`s
because we have no key of that name anywhere that I can find

Artefact2:
It used to be, see https://github.com/kmonad/kmonad/blob/7b5142b85660bec043f323ab32c4240a30332473/src/KMonad/Keyboard/Keycode.hs#L352

It seems a lot of the old abbreviations did not make it through in the new code.

slotThe:
Ah I see.  Anything else in particular you are also missing?

MuhammedZakir:
@slotThe is right.

I presume `pp` is referring to `play/pause` button? Currently, those aliases are not defined [1]. It was commented out in `keycode-refactor` branch, and @david-janssen probably forgot to re-add them.

[1] https://github.com/kmonad/kmonad/blob/2356e2734c029e9cf6af7781afce706fa00e8fa7/src/KMonad/App/Parser/Keycode.hs#L121

---

@david-janssen: Temporarily, we can at least add aliases for keycodes that are present in all 3 OSs. I did a quick-look, and the only problematic aliases are for `wakeup`, `launchpad` and `mission_control`.

Artefact2:
Where can I find a list of all the "core names" ? I think a lot were left out, considering that the [typematrix template](https://github.com/kmonad/kmonad/blob/master/keymap/template/type_matrix_2030.kbd) badly fails too. (It has a few exotic keycodes like katakanahiragana, ro, etc.)

[...]

Issue #365, opened 2021-09-28
Apple magic keyboard on linux


jakubdibala:
As described in https://github.com/kmonad/kmonad/issues/72, i tried to patch kmonad to make **FN** key working, but upstream has changed so I cannot patch it. I'm not able to make these edits in Haskell 😞 

Is there any solution to use FN key on Apple Magic Keyboard? As soon as I press FN, kmonad crashes. My current workaround is to run kmonad as systemd service, which is always restarting after crash. But I need to remap FN to `lctr`.

Any ideas? Thanks! 🙂

david-janssen:
Hey there, I had a look and that pull request was actually based on a complete refactoring of the keycode stuff that I was doing, because the keycode encoding was still from a time we only supported Linux. We've just redone all this stuff, and I need to rebase our dev branch onto master, so I hope to have a solution for you soon (ish). I'll get started on the rebasing now and keep you up to date.

jakubdibala:
@david-janssen Oh, ok, thank you for explaining. I'll wait for update 🙂

slotThe:
@jakubdibala you can track the progress in #372

Issue #72, opened 2020-09-08
[Linux] Bug: using fn key on apple keyboard crashes kmonad


thoelze1:
I'm running a fresh build of `master` under Debian on a MacBookPro12,1 with `-l debug`. Pressing <kbd>fn</kbd> on the builtin keyboard crashes kmonad with the following output:

```
Closing process: emitter_proc
Closing KeySource
Releasing ioctl grab
Closing KeySink
Unregistering Uinput device
Closing Uinput device file
kmonad: toEnum{Keycode}: tag (464) is outside of enumeration's range (0,255)
CallStack (from HasCallStack):
  error, called at src/KMonad/Keyboard/Keycode.hs:312:32 in kmonad-0.4-I8ZsE4oWg3dF5zI18fnym4:KMonad.Keyboard.Keycode
```

Has anyone else experienced this issue yet? Maybe @eighal?

eighal:
> Maybe @eighal?

Yes, same here.  (It prints that call stack for me even without `-l debug`.)

david-janssen:
So, looking at that it seems to indicate to me that the Macbook keyboard hardware is sending a key-event to KMonad that we had never anticipated, i.e. with an unparseable keycode.

I'm guessing `evtest` would also show some kind of keycode that is above 255?

The solution would probably be to add support for arbitrary int-values above the expected keycodes, perhaps define an event's keycode value as 'Either Keycode Int', (the alternative: expanding the Keycode to cover this 464 value, would mean losing correspondence between Keycode and 'real' linux-kernel keycode values.

A bit too tired to come up with the optimal solution right now, but if you guys can gather a bit more data, it sounds like a clear and solveable problem.

david-janssen:
Of course, the ultimate upside of this is that we should probably be able to rebind the Fn key now :-)

eighal:
Yes!  Since it is a small keyboard (although no Atreus), that is really good news.

eighal:
> I'm guessing `evtest` would also show some kind of keycode that is above 255?

Yes: I get 464.

david-janssen:
Very good, I'll need to do a bit of refactoring of the core of KMonad to make this work, we need to change how we represent keycodes. Luckily: this is Haskell :-). Super busy atm, so might not get to it soon though. Is there a way for you to ignore that key? Otherwise it might be quite easy to just hack an 'ignore-crash' thing into keycodes as a temporary duct-tape fix.

eighal:
[...]

PR #372, opened 2021-10-05
Keycode refactor

david-janssen:
Alea no iacta est


david-janssen:
As mentioned in #371 this update breaks a lot of configs because it doesn't support all of the aliases.

Solution is simple: add the aliases back in, is quite straightforward.

Additionally, add a unit-test that runs the parser over all user-provided configurations and checks that the parser never fails. Not a full coverage test, but better than what we have.

MuhammedZakir:
Some keycodes such as `volume up/down`, `mute`, `play/pause`, etc. are missing from Windows keycodes.

david-janssen:
> Re-posting: Some keycodes such as volume up/down, mute, play/pause, etc. are missing from Windows keycodes.

Ahhh, I originally defined the 'CoreNames' as a collection of all the keys on a standard 100% keyboard. Maybe there are other keys that are present in all OSes that we also want to support?

So, keeping track of things:
- Volume control
- Song-play-pause
 
There are keycodes for this in Mac?

david-janssen:
Gonna have to put this down for the rest of today and tomorrow, master is in a stable state, and we know what needs to be tinkered on for keycode-refactor.

The way I see things is: now is our opportunity to really get the keycode stuff nailed down. Ideally we: get an overview of all abstract keycodes supported by mac/windows/linux. Then the intersect of those should map to some set of CoreNames. Each corename assigns 1 name to each keycode in this set (the mapping will differ between OSes, and of course, so will the underlying 'Keycode' type).

On top of those core names we should then define a bunch of aliases, so that people can use shorthand or longhand to refer to the same keycode.

Finally, we need some sort of way of reading literal keycodes from the config-language. Perhaps something along the lines of `&123` to describe keycode `123`. I would like this to be the same in Mac, so that a config-file will parse, regardless of the underlying OS. I know keycodes in Mac are 2 numbers, so maybe we glue them together?

MuhammedZakir:
> > Re-posting: Some keycodes such as volume up/down, mute, play/pause, etc. are missing from Windows keycodes.
> 
> Ahhh, I originally defined the 'CoreNames' as a collection of all the keys on a standard 100% keyboard. Maybe there are other keys that are present in all OSes that we also want to support?

I guess so.

> So, keeping track of things:
> 
> * Volume control
> * Song-play-pause
> 
> There are keycodes for this in Mac?
:
Yes. In Mac, there are keycodes for all those commented out `aliases` except for `sleep` and perhaps, `print` or `forward`.

---

> The way I see things is: now is our opportunity to really get the keycode stuff nailed down. Ideally we: get an overview of all abstract keycodes supported by mac/windows/linux. Then the intersect of those should map to some set of CoreNames. Each corename assigns 1 name to each keycode in this set (the mapping will differ between OSes, and of course, so will the underlying 'Keycode' type).

First of all, I suggest to create a file (markdown?) and add all keycodes we would support there. there should be two sections - one for corenames and one for aliases, and this should be further split into two sections - one for common keys and one fore OS-specific keys. This can be considered as a refernce manual and will really make things easier for both of us - devs and users.

> On top of those core names we should then define a bunch of aliases, so that people can use shorthand or longhand to refer to the same keycode.

How to handle OS-specific aliases?

> Finally, we need some sort of way of reading literal keycodes from the config-language. Perhaps something along the lines of &123 to describe keycode 123. I would like this to be the same in Mac, so that a config-file will parse, regardless of the underlying OS. I know keycodes in Mac are 2 numbers, so maybe we glue them together?

A seperator is needed. Otherwise it will be hard? Keycode is in the form of `usage page` + `usage` and supported `usage page`s are - `0x7`, `0xC`, `0xFF` and `0xFF01`. `0xFF01` in decimal have 5 digits, so in Mac, parse first five digits as `usage page` and rest as `usage`? This **must** be documented if we're going to do this way..

MuhammedZakir:
> First of all, I suggest to create a file (markdown?) and add all keycodes we would support there. There should be two sections - one for corenames and one for aliases, and this should be further split into two sections - one for common keys and one fore OS-specific keys. This can be considered as a refernce manual and will really make things easier for both of us - devs and users.

Where can I find keycodes and their name for Windows? With that, I can write this.

P.S. After making this, it will be trivial to cross-check keycodes.

rwese:
> Where can I find keycodes and their name for Windows? With that, I can write reference/summary of all (supported) keycodes of all platforms.

This should be what you are looking for:

* https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
* http://kbdedit.com/manual/low_level_vk_list.html

hope that helps.

david-janssen:
@MuhammedZakir Not sure what you are working on or have been working on these last 22 hours, just popping in to say that I am going to make that reference document now. Trying to avoid duplicate effort :)

david-janssen:
Just pushed my first draft, realizing just how much work this is. Exhausted from being sick and travelling. Gonna put this down for a few days.

https://github.com/kmonad/kmonad/blob/keycode-refactor/doc/keycode_reference.org

david-janssen:
Dude... you did a lot :)

Well done :1st_place_medal:

david-janssen:
Actually, slept on it a night, and looking at it again with fresh eyes, I think "getting it merged without breaking configs" is going to be a real headache. I'll basically have to reimplement the old parsing algorithm and add deprecation warnings, and then have to add all the OS-remappings by hand, so that for a few weeks it'll work, and then we'll delete all that stuff again.

I think I'm just going to do the 'git gud' bit mentioned above, and then we just do a breaking version change, with our sincerest apologies for lack of warning :).

I'm just going to keep plugging away at this. Somehow I feel like I should pick a target and just stop thinking and move towards it, that seems like how I have been encouraged to work before. But I really enjoy changing my mind the whole way through until something is shiny enough to actually manifest long enough to come to fruition. So um... yah. Onwards and upwards and such.

MuhammedZakir:
> Hey Muhammed, I have a question: how does Mac signal repeat events? I.e. if you press some key then you get a press event, but what does Mac do internally when you hold that key? Does it just send a series of presses? Does it alternate between presses and releases?

When a key is held down, it sends a series of presses.

MuhammedZakir:
Do you need any help? Feel free to ping me when/if there is something I can do. :-)

david-janssen:
@MuhammedZakir 

Hey there, thanks for the update and the info about Mac key-repeat behavior :-). Things have been extraordinarily bad here the last few days, illness wise, until about yesterday afternoon, so sorry for the silence. 

I've managed to start sketching out the problem very carefully, and I think I'm starting to get somewhere. Not sure there are many ways for you to help right now, I've gotten into a state where I've jumbled everything up, and now I need to unjumble it again 'correctly'. I'll keep you in the loop when I make progress.

In regards to your questions:

> I hope you won't remove already added corenames.

The intention is: basically no. I want to simplify the way we deal with keynames by using only 1 name per 'semantic' key in KMonad. So no more 'LeftShift' 'lsft' 'lshft' management. I'll just pick one. However, I will create the ability for people to include their own aliases. We manage 1, but anyone can specify anything they want.

> It won't work because some "common" keys are OS-specific. For instance, Mac has spotlight button which isn't in other OSes.

Actually, I'm not sure that's true. If you install Windows on a Mac, and then you press the spotlight button, then either:
- Windows never sends an event at all, which means we'll never be able to catch it
- Windows sends an event with some keycode

The real issue is: figuring out what numbers correspond to what keys. That's why making keynaming extensible might be a good idea. We can't figure out what spotlight does on windows, but if we make it discoverable, then:
1. People can easily fix the problem on the spot
2. People can contribute their findings without having to touch Haskell
3. We might even get support for different locales out of the box.

> It is kind of a hacky. 

Yah, that is kind of hacky :-). I think the 'pure' problem of: if I press a mac button in Windows, *something* must happen, let's model that problem as cleanly as possible, and then hack *around* it is probably a better longterm solution. I think I'm trying to get the hacks out of keycode.

However: support for OS-specific functionality for buttons is a cool idea and something that we should probably include. Especially because otherwise command-buttons wouldn't be cross-platform. So we definitely need something like:

```
(defalias destroy-everything
  :mac  (command-button "rm  -rf /Users/my-name/*"
  :linux (command-button "rm -rf /home/my-name/*"
  :windows (command-button "del /Users/my-name/*"
)
```

MuhammedZakir:
Error when executing `kmonad` from `tinker` branch:

```
kmonad: /home/david/prj/kmonad/keycode_table/en_US.md: openFile: does not exist (No such file or directory)
```

david-janssen:
Popping in again to say that I managed to do that think I was saying: define configurable settings in 1 place and then hook up both the parser and the invocation. If interested, check the Tinker branch and inspect the following 3 modules

KMonad.App.Configurable <- definitions of 'things that can be configured'
KMonad.App.Parser.Grokker <- use configurables to very easily parse settings from a config file
KMonad.App.Invocation.Parser <- use configurables to very easily generate cmd-line interface

All of this stuff is built ontop of the keycode stuff.

I still need to finish the grokker and hook it up to the model and do a bit of tweaking on the Run modules, but then I will be approaching the point of release.

This is going to be a breaking release. People will need to make a few edits to their config file, and KMonad will be invoked a little bit differently. Maybe a... 10% change? But it will come with some new features and a much more solid core, ready to start implementing some of the changes that I've been putting off.

Anyways, onwards and upwards.

MuhammedZakir:
Note to self:

Mac keycodes for `zenk` and `hang` found - https://gist.github.com/MightyPork/6da26e382a7ad91b5496ee55fdc73db2#file-usb_hid_keys-h-L179-L192.

https://github.com/kmonad/kmonad/blob/1d2276069fd781c4518082f1b36f3b2880667e79/src/KMonad/App/Parser/Keycode.hs#L102-L104

---

What exactly is that `documents` key (line 102)? Is that really a common key? A quick search didn't yield any result for both Windows and Mac [1]. Maybe it's better to remove/comment it to make building on those OSes working?

[1] Actually, for Linux too; probably because of the name "documents".

david-janssen:
So, I have been gone for a while. Part of me wants to apologize and explain, but without getting into much detail. So, here goes: illness, medication-side-effects, stress. I really wish I was more reliable. There, I hope that suffices.

Now, onwards with the task at hand: the keycode refactor has gotten out of hand for me. It is so big that I can't fit it into my head, moreso after all these months of absence. Last week I already put key-repeat for Windows into master. This was one of the improvements made in `keycode-refactor`.

I want to delete this branch, but there is a lot of valuable work that was done here. So, for anyone using this branch right now, could you please list the reasons? What are the concrete improvements in `keycode-refactor` that make you use it over master? 

The reason behind this inquiry: my mind simply doesn't manage to hold such large mouthfuls of complexity lately, and in trying to I keep throwing up (mentally). If we could break this task up into small, chewable mouthfuls of complexity, I think I can make progress on merging these things into master. 

I am going to ping @MuhammedZakir because I am sure he uses `keycode-refactor`. @MuhammedZakir could you also ping anyone else that you know who is using this branch?

MuhammedZakir:
> I am going to ping @MuhammedZakir because I am sure he uses keycode-refactor. @MuhammedZakir could you also ping anyone else that you know who is using this branch?

Basically
1. this has better key support.
2. because of #_1_, this doesn't has random crashes. On `master` branch, pressing certain keys crashed KMonad.

MuhammedZakir:
Scancodes in different OSes (may help in finding common keys):
1. https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values
2. https://github.com/houmain/keymapper/blob/main/src/runtime/Key.h

ZerdoX-x:
Sorry guys, can't we implement any workaround for fn button? It's not that big issue when it can be fixed with single line of code as in #475. Like.. Do we really need to rewrite the whole project for 2 years?

There are still issues like #72 and #365. I mean.. If I have important functionality on fn button (mine opens apps launcher) then kmonad is fully unusable because process just crashes, and afaik there is no way to ignore fn key while I am describing `defsrc`

I've spent about an hour compiling, learning and configuring kmonad and I guess I'll spend one more hour patching & compiling kmonad again :(

ZerdoX-x:
ready! 🥳

jokesper:
> ready! 🥳

Do you refer to the `fn` issue. Since this was fixed via #935.

PR #379, opened 2021-10-09
optional :timeout-button for tap-hold-next{,-release}

### LINKED ISSUE #378: defaulting to tap after a timeout
Here's what I'm trying to achieve:

1. tap spacebar: type a single space `_`
2. hold spacebar while typing letters `a` `b` `c`: type uppercase letters `ABC`
3. hold spacebar long enough without typing anything else: type a bunch of spaces `______`

Note that in case 3, the repetition is achieved via the operating system's key-repeat functionality; KMonad would simply emit a `spc` press event as normal. I'm not trying to emulate key-repeat inside kmonad, so this is not a duplicate of #290.

I am able to achieve cases 1 and 2 easily using `(tap-hold-next-release 500 spc sft)`, but after 500ms this button defaults to `sft` instead of `spc`.

I'm currently thinking of adding an optional `:timeout-button` argument to `tap-hold-next-release` to specify which button (same as the tap button, same as the hold button, or a third completely different button) to use when the timeout expires, does that sound like a good way to do it?

gelisam:
If the `:timeout-button` argument is specified, `tap-hold-next-release` will switch to that button instead of the hold button if the timeout elapses. This makes it possible to make implement a button which can both be tapped and pressed-and-held in order to type 1 or multiple (via the operating system's auto-repeat functionality) copies of a character, while still behaving as a modifier key when pressed in combination with other keys.

Fixes #378.

### DIFF
--- a/src/KMonad/Args/Joiner.hs
+++ b/src/KMonad/Args/Joiner.hs
@@ -368,7 +368,8 @@ joinButton ns als =
     KAround o i        -> jst $ around             <$> go o <*> go i
     KTapNext t h       -> jst $ tapNext            <$> go t <*> go h
     KTapHold s t h     -> jst $ tapHold (fi s)     <$> go t <*> go h
-    KTapHoldNext s t h -> jst $ tapHoldNext (fi s) <$> go t <*> go h
+    KTapHoldNext s t h mtb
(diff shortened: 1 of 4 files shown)


slotThe (approved):
This is really cool, thanks!  I will merge this in a few days if no one objects

gelisam:
oh wait I forgot to edit the documentation and the changelog

PR #402, opened 2021-11-07
Add Ú compose sequence

jdnavarro:
These already come with X11, or at least I can see them in my `/usr/share/X11/locale/en_US.UTF-8/Compose`.

In Spanish they are extremely rare but still valid.

### DIFF
--- a/src/KMonad/Keyboard/ComposeSeq.hs
+++ b/src/KMonad/Keyboard/ComposeSeq.hs
@@ -148,6 +148,7 @@ ssComposed =
     , ("~ O"      , 'Õ'     , "Otilde")
     , ("\" O"     , 'Ö'     , "Odiaeresis")
     , ("\" U"     , 'Ü'     , "Udiaeresis")
+    , ("' U"      , 'Ú'     , "Uacute")
     , ("x x"      , '×'     , "multiply")
     , ("/ O"      , 'Ø'     , "Oslash")
     , ("' Y"      , 'Ý'     , "Yacute")


src/KMonad/Keyboard/ComposeSeq.hs
```diff
@@ -178,8 +178,10 @@ ssComposed =
     , ("/ o"      , 'ø'     , "oslash")
     , ("` u"      , 'ù'     , "ugrave")
     , ("' u"      , 'ú'     , "uacute")
+    , ("' U"      , 'Ú'     , "Uacute")
     , ("^ u"      , 'û'     , "ucircumflex")
     , ("\" u"     , 'ü'     , "udiaeresis")
+    , ("\" U"     , 'Ü'     , "Udiaeresis")
```
slotThe, line 184: This already seems to be defined [here](https://github.com/kmonad/kmonad/blob/64aaba001a535ec4cd49525eb658823468d6d58d/src/KMonad/Keyboard/ComposeSeq.hs#L150)

jdnavarro: I removed it. It was missing in 0.4.1, but didn't realize it was already in `master`.

slotThe: ah yeah seems to have been added in https://github.com/kmonad/kmonad/commit/b7214c4bfa7ae2d6d8ea99fd991aeb3b951fdb5a

slotThe: @jdnavarro doesn't seem like it's been removed :)

jdnavarro: Sorry about that, there you go.

slotThe: Perfect, thanks!

PR #434, opened 2021-12-05
Update installation.md macOS dext package

quatauta:
Update doc/installation.md to highlight the pre-build and signed
dext package in the dext git submodule.

It took me quite a while to notice that the dext extension is already
provided by the dext submodule.


slotThe:
Thanks!

PR #442, opened 2021-12-16
typo: /etv/ -> /etc/

NiloCK:
...for the very small number of people who might be reading this doc but not able recognize & correct this error.


slotThe:
Thank you!

PR #452, opened 2022-01-01
Add layout for T430

Anskrevy:
The T430(s) has a pretty cramped top row. Maybe this can help someone with formatting their own layout. The base for this was the x200 layout already in this repo.


startup/uinput.conf
```diff
@@ -0,0 +1,22 @@
...
+# When the exit code is 0/SUCCESS it means the module has been successfully
+# inserted. This can be confirmed by running:
+
+# lsmod | grep uinput
+
+# Credit and a more thorough explanation for auto loading a module checkout
+# this stack exchange thread:
+# https://unix.stackexchange.com/questions/71064/systemd-automate-modprobe-command-at-boot-time
+
+uinput
```
slotThe, line 22: > Add uinput.conf to startup files. For systemd users uinput.conf can be used to auto load the uinput module. There is already a file to help for creating a systemd service for Kmonad but without auto loading that module this kind of defeats the purpose of having kmonad start at boot. The file tells a user where to place it and how to verify it is working.

This sounds like weird behaviour to me—I don't think having to manually start `uinput` is very common (at least I have never heard users complain about this, which I'm sure would have happened)

slotThe:
> This is my first pull request so I'm sorry that this is getting a little messy. I wanted to submit 2 requests but I guess any changes I make in master get updated here.

You would normally have two different branches for this, then you can submit one pull request from, e.g., the `add-t430` branch and another one from the `systemd-uinput` branch

Anskrevy:
> You would normally have two different branches for this, then you can submit one pull request from, e.g., the `add-t430` branch and another one from the `systemd-uinput` branch

That makes sense. Well I went ahead and deleted the uinput.conf. Maybe something is wrong with my system but I had to run `modprobe uniput` after my first restart when I got everything working. I've removed that so now this is just a pull request for the keymap. Largely it works with a lot of the T4XX series after the T430, except not every keyboard has the first line
```lisp
       mute vold volu
```
but everything after that should work. Maybe a note in the file would help point this out; as well as, renaming the file to something like `thinkpad_T430-T4X0_iso.kbd` where `X` is the last in the series that are compatible with this layout (I think it may be the T490 but I don't know for certain). Or just merge as is, whatever seems best to you is fine by me.

slotThe:
> but everything after that should work. Maybe a note in the file would help point this out; as well as, renaming the file to something like `thinkpad_T430-T4X0_iso.kbd` where `X` is the last in the series that are compatible with this layout (I think it may be the T490 but I don't know for certain). Or just merge as is, whatever seems best to you is fine by me.

I think people won't have much trouble adjusting this to their needs if they use a newer model in the T-series; merging as-is sounds fine to me.

I just rebased on top of `master` (instead of merging) and squashed everything down to one commit (since, at this point, we're only doing a single thing)—thanks!

PR #464, opened 2022-01-15
warninng to default disabled service on startup

jhonasn:
My void linux created this down file by default disabling the kmonad service (created with xmksv) and took me some days to note this, so I want to warn newcomers to not spend so much time on this as I did.


slotThe:
Thanks!

PR #465, opened 2022-01-16
Minor fix to installation.md

fabionatali:
The documentation refers to `$(guix build kmonad)` as a way to get the exact
KMonad path in a Guix package store.

The build command, however, does not get expanded to the intended path
`/gnu/store/<hash>-kmonad-0.4.1/`. Instead, it returns the store paths of all
the package dependencies, e.g. including
`/gnu/store/<hash>-kmonad-0.4.1-static/`.

Therefore, the following copy operation, as currently mentioned in the docs, is
broken.

```
cp $(guix build kmonad)/lib/udev/rules.d/70-kmonad.rules /lib/udev/rules.d/
```

A simple fix is proposed in this commit, that indicates how to identify the
correct path for the copy operation.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -288,15 +288,40 @@ Kmonad is available in the Arch User Repository (AUR) as
 
 ### GNU Guix
 
-You can install `kmonad` via the `guix` package manager. you will need to copy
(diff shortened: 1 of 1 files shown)


slotThe:
Thank you!

PR #496, opened 2022-03-04
Update deflayer in iso_tkl.kbd to match defsrc

ego-cv:
Deflayer had 85 keys, missing 3 keys that are in the defsrc.


slotThe:
Thanks!

PR #504, opened 2022-03-14
Add a note regarding dext activation failure

MuhammedZakir:
Suggest `forceActivate` if using `activate` failed.

P.S. This should also be added in `keycode-refactor`. When merging, could you also update it? If not, I can open a new PR.


slotThe:
Thank you!

> P.S. This should also be added in `keycode-refactor`. When merging, could you also update it? If not, I can open a new PR.

We will rebase `keycode-refactor` before merging it into master (or at least cherry pick commits like this one), so this shouldn't be a problem

Issue #489, opened 2022-02-22
[Windows] Printscreen doesn't work when kmonad is running


6gk:
Quick example config:

```
(defcfg
	input  (low-level-hook)
	output (send-event-sink)

	fallthrough   true
	allow-cmd     true
)

(defsrc         esc grv)
(deflayer base  grv esc)
```

Running kmonad via a shortcut with the Target of `C:\Programs\kmonad.exe C:\Programs\kmonad.kbd` and the start directory being `C:\Programs`

Version: `v0.4.1` (latest release, i don't want to compile on windows)

When using [autohotkey as a subpar alternative to xev](https://superuser.com/questions/348070/equivalent-for-xev-on-windows), it shows `2A  000	a	d	1.91	not found` when pressing the key, as opposed to `2C  137	 	d	1.00	PrintScreen` (`VK  SC  Type  Up/Dn  Elapsed  Key`).

MuhammedZakir:
Pass `--log debug` cli option to `kmonad.exe` and post the output you get when you press `PrintScreen` here.

6gk:
![image](https://user-images.githubusercontent.com/40295114/155208518-a57d2b6d-815c-4016-8c1e-b53fc3953b6e.png)

MuhammedZakir:
Apparently, there are two "print" keys and unfortunately, both of them were given the same name in KMonad. We will have to update the code to fix this.

https://github.com/kmonad/kmonad/blob/2cc865bcfc255e16225f1980d214188a99205297/src/KMonad/Keyboard/IO/Windows/Types.hs#L165-L167

https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes:
```
VK_PRINT	0x2A	PRINT key
VK_EXECUTE	0x2B	EXECUTE key
VK_SNAPSHOT	0x2C	PRINT SCREEN key
```

6gk:
Whoops, sorry, didn't see that email!

> Edit: What are the values in second column?

I should probably have formatted it differently, I'm pretty sure it was:
```
VK SC  Type Up/Dn Elapsed Key
2A 000 a    d     1.91    not found  # pressing when kmonad is running
2C 137      d     1.00    PrintScreen  # pressing when kmonad isn't running
```

> And also why does it show "not found" for 0x2A? (note: I haven't used AHK.)

[...]

PR #505, opened 2022-03-15
Windows: make PrintScreen work

MuhammedZakir:
There are two "print" keys in Windows - a legacy (unused?)
PrintScreen key and the currently common PrintScreen key.
Unfortunately, KMonad uses same name for both of them
and only exposes the legacy print key. This fixes it by
renaming the legacy one as `KeyPrintLegacy` and exposing
it as `prntl` and exposing `PrintScreen` as `prnt` for
consistency across OSes.

Fixes: https://github.com/kmonad/kmonad/issues/489

### DIFF
--- a/src/KMonad/Keyboard/IO/Windows/Types.hs
+++ b/src/KMonad/Keyboard/IO/Windows/Types.hs
@@ -162,9 +162,9 @@ winCodeToKeyCode = M.fromList $
   , (0x27, KeyRight)
   , (0x28, KeyDown)
   -- , (0x29, ???)            -- Defined as VK_SELECT
-  , (0x2A, KeyPrint)
+  -- , (0x2A, ???)            -- Defined as VK_PRINT (legacy PrintScreen)
   -- , (0x2B, ???)            -- Defined as VK_EXECUTE
-  , (0x2C, KeyPrint)          -- Defined as VK_PRINT_SCREEN
+  , (0x2C, KeyPrint)          -- Defined as VK_PRINT_SCREEN / VK_SNAPSHOT
   , (0x2D, KeyInsert)
   , (0x2E, KeyDelete)
   , (0x2F, KeyHelp)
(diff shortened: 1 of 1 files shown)


MuhammedZakir:
What I am not sure about is whether to keep lagacy print key. Does anyone use that key? Should we remove it and add it only when a user asks for it?

src/KMonad/Keyboard/Keycode.hs
```diff
@@ -305,6 +305,9 @@ data Keycode
   | KeyFn
   | KeyLaunchpad
   | KeyMissionCtrl
+#endif
+#ifdef mingw32_HOST_OS
+  | KeyPrintLegacy
```
slotThe, line 310: > What I am not sure about is whether to keep lagacy print key. Does anyone use that key? Should we remove it and add it only when a user asks for it?

I'd say the less `CPP` we use the better, but it's already enabled for this file anyways so...  I don't really have an opinion, as I'm not a Windows user.  I'll go with whatever people who use it say would be the best option

slotThe:
@6gk @whitbur would any of you be willing to test this/offer an opinion as to how we should handle `KeyPrintLegacy`?

6gk:
I'll test this version tomorrow (not on windows + about to go to bed), but [the commit Muhammed posted in the issue I made](https://github.com/kmonad/kmonad/issues/489#issuecomment-1059907321) solved my issue flawlessly (commits appear similar, but I'm half-asleep).

And I'ven't a clue about KeyPrintLegacy -- The key does nothing for me, which leads me to doubt there's any use for it in modern Windows. ​ Removing it *shouldn't* bring any harm, but same with keeping it. (I might've misunderstood you though?)

whitbur:
I tested this on Windows 10, validated that taking a screenshot using the PrintScreen key before did not work and that with this commit I am able to take a screenshot. Looks good to me.

On `KeyPrintLegacy` my opinion's the same as it was on #489. I don't see any reason to keep it around, I can't imagine anyone is using it. Personally I don't think it's worth the minor additional complexity to support it, so I would remove it from the commit. I'd still support shipping this PR as is if others do feel it's important, though.

slotThe:
Thanks to you two!  @MuhammedZakir I'd prefer the cleaner approach then—let's remove `KeyPrintLegacy` and see if anyone complains :)

MuhammedZakir:
@slotThe: Done.

slotThe:
Great, thank you!

PR #509, opened 2022-03-23
Keymap for X220/X230 Japanese keyboard - using the japanese keys as t…

lapidot:
…humb keys for Numpad and Navigation Layers

The Japanese keyboard adds 3 thumb keys that can be used as a thumb cluster. Used them for Nav and Numpad layers, also experimenting with separating the hands by 1u (mostly for a more natural thumb position on an extra thumb key)
The thumb keys act as space when tapped and as layer changers when held.
Love it!


slotThe:
Thanks!

PR #513, opened 2022-03-28
Removed a redundant bracket from an example on line 816 in tutorial.kbd

B1T3X:
I guess you have to start somewhere, so this PR introduces the silliest, smallet change ever.
I have noticed an extra bracket on an example on line 816 so I removed it.


keymap/tutorial.kbd
```diff
@@ -813,7 +813,7 @@
   -------------------------------------------------------------------------- |#
 
 (defalias
-  mt  (multi-tap 300 a 300 b 300 c 300 d e))
+  mt  (multi-tap 300 a 300 b 300 c 300 d e)
```
MuhammedZakir, line 818: That is the corresponding closing bracket of the one before `defalias`. This is a valid config and is usually written like that by lispers. If it really trips you, you can move that closing bracket to the next line.

PR #516, opened 2022-03-29
[Mac] Support more identifiers for filtering devices

MuhammedZakir:
- Mac: Support more identifiers for filtering devices
- Mac: list-keyboards: Print all available identifiers
- Mac: Change functions to accept `DeviceIdentifiers` record
- Something's wrong. Add tracing function.
- Mac: Fix for compilation errors.

**For some reason, this is not working when a product name is given in `iokit-name`. Need help!**

cc: @slotThe @thoelze1

### DIFF
--- a/src/KMonad/App/KeyIO/Common/Types.hs
+++ b/src/KMonad/App/KeyIO/Common/Types.hs
@@ -71,14 +71,23 @@ instance Default UinputCfg where
 -- $cfgs-mac
 
 -- | Holds the product string of keyboard if given. Else, Nothing.
-newtype IOKitCfg = IOKitCfg
-  { _productStr :: Maybe Text
+data IOKitCfg = IOKitCfg
+  -- Prepend `cfg` to distinguish from `DeviceProperties`'s fields.
+  { _cfgManufacturer :: Text
+  , _cfgProductName  :: Text
+  , _cfgSerialNumber :: Text
+  , _cfgTransport    :: Text
+  , _cfgCountryCode   :: Word64
+  , _cfgLocationID    :: Word64
+  , _cfgProductID     :: Word64
+  , _cfgVendorID      :: Word64
+  , _cfgVersionNumber :: Word64
   } deriving Show
 makeClassy ''IOKitCfg
 
 -- | By default, no product string (will grab all keyboards).
 instance Default IOKitCfg where
-  def = IOKitCfg Nothing
+  def = IOKitCfg "" "" "" "" 0 0 0 0 0
 
 -- | No config options for Kext/Dext
 data ExtCfg = ExtCfg deriving Show
@@ -107,7 +116,7 @@ instance HasKeyRepeatCfg SendEventCfg where keyRepeatCfg = seRepCfg
 
 data KeyInputCfg
   = LinuxEvdevCfg          EvdevCfg
-  | MacIOKitCfg            IOKitCfg
+  | MacIOKitCfg            (Maybe IOKitCfg)
   | WindowsLowLevelHookCfg LowLevelHookCfg
   deriving Show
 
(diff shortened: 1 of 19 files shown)


src/KMonad/App/KeyIO/Common/Types.hs
```diff
@@ -107,7 +116,7 @@ instance HasKeyRepeatCfg SendEventCfg where keyRepeatCfg = seRepCfg
 
 data KeyInputCfg
   = LinuxEvdevCfg          EvdevCfg
-  | MacIOKitCfg            IOKitCfg
+  | MacIOKitCfg            (Maybe IOKitCfg)
```
MuhammedZakir, line 119: Is this okay?

src/KMonad/App/KeyIO/Mac/IOKitSource.hs
```diff
@@ -53,30 +97,50 @@ waitKey ptr = do
...
-          Just s -> do
-            logInfo $ "Opening HID device: " <> s
-            liftIO $ withCString (unpack s) grabKb
-
+          (Just cfg) -> do
+            liftIO $ do
+              dp <- deviceProperties cfg
+              str <- print dp
+              let t f dp = (D.trace ("DeviceProperites: \n" <> str) f dp)
+              t grabKb dp
```
MuhammedZakir, line 136: Note to self: remove this.

src/KMonad/App/KeyIO/Mac/IOKitSource.hs
```diff
@@ -86,3 +150,21 @@ withIOKitSource cfg = mkCtx $ \f -> do
                            (Keycode $ re^.reCode)
 
   bracket init cleanup $ \ptr -> f (nextEvent ptr)
+
+  where
+    print dp = do
+      dp <- peek dp
```
MuhammedZakir, line 156: Note to self: remove this.

Can this `where` section be moved to below `init` function? I am getting errors when I do that. I guess I am still not clear on where I can use `where`.

src/KMonad/App/Parser/Tokenizer.hs
```diff
@@ -333,7 +333,9 @@ itokens :: [(Text, Parser IToken)]
 itokens =
   [ ("device-file"   , KDeviceSource <$> (T.unpack <$> textP))
   , ("low-level-hook", pure KLowLevelHookSource)
-  , ("iokit-name"    , KIOKitSource <$> optional textP)]
+  , ("iokit-name"    , KIOKitSource <$> optional (textP >>= mkCfg))]
+  where
+    mkCfg n = pure $ def {_cfgProductName = n}
```
MuhammedZakir, line 339: This is for testing only. Will write a proper parser after we get this working.

slotThe:
@MuhammedZakir: https://github.com/kmonad/kmonad-dext

You should have an invite for the right perms

MuhammedZakir:
I didn't forget this. It will take a few more weeks before I am free to work on this. Sorry!

Before getting into this, I would like to know what you guys think about the question I posted above. Should we support capturing multiple devices using identifiers? I think it maybe useful for some split-keyboards.

david-janssen:
> I didn't forget this. It will take a few more weeks before I am free to work on this. Sorry!

No worries, I can relate... ;-) On the same note let me apologize for being gone for a while too. Life fluctuates and we do what we can, and no more.

> Should we support capturing multiple devices using identifiers? I think it maybe useful for some split-keyboards.

I think this is a good idea. Ideally we would have a stream of uniquely identified keys. In the case of multiple keyboards, that means keyboard-id and keycode-id. I am again starting to think about how to merge ideas from keycode refactor into master, and I will be keeping this in mind as I work on the KeyEvent representations.

kieran-mace:
This may or may not be useful for your PR, but this comment seems to be affecting M2 Macs from identifying keyboards: https://github.com/kmonad/kmonad/issues/633#issuecomment-1326744388

Issue #256, opened 2021-05-16
tap-next-release and order of release events


jchtt:
I mentioned this particular problem before in #111, see [here](https://github.com/kmonad/kmonad/issues/111#issuecomment-716747605). When using `tap-next-release` (and `tap-hold-next-release`), key release events of other keys are emitted immediately, while I think it would be preferable to put them on hold until the `tap-next-release` key is resolved.

For example, consider this configuration:
```
(defsrc a)
(deflayer default @act)
(defalias
  act (tap-next-release b lctl)
)
```
When inputting `Pshift Pa Rshift Ra`, kmonad currently translates this into `Pshift Rshift Pb Rb` (a lowercase `b`). Since I have some `tap-hold-next-release` buttons on my home row, this behavior often leads to typos when I want to type the associated letters in uppercase since I am not holding the shift key long enough to finish the release of the other key.

I think it would be better if the above got translated into `Pshift Pb Rshift Rb` (an uppercase `B`), keeping the original order of key presses and releases intact. This also seems overall more predictable for the user.

thoelze1:
I agree that `tap-next-release` (and `tap-hold-next-release`) should behave as you describe.

 I couldn't figure out how to fix this on `master`. The `tap-next-release` button in KMonad.Model.Button pauses processing of other events with `hold True` and later unpauses it with `hold False`. It appears that `hold` only pauses processing of as-yet-unpressed buttons, and so it does not (cannot?) pause the release of `shift` (because it's already been pressed). Maybe there is a simple fix on `master` that I don't know how to implement, or maybe this will be easier to change in `keycode-refactor`. Any insight @david-janssen?

david-janssen:
> Maybe there is a simple fix on master that I don't know how to implement, or maybe this will be easier to change in keycode-refactor. 

Sadly neither, however, that is why I am already dreaming of `model-refactor` :)

PR #524, opened 2022-04-18
Change release button handling to issue passthrough events

jchtt:
Hi all,

this is meant to address #256, release events not being held when using `tap-next-release`. From studying the code, this seems to be the least invasive solution to this problem.

The issue stems from the way KMonad handles key releases in relation to the pull chain. The way I understand it, the pull chain consists of dispatch, hooks, and sluice. Dispatch listens to incoming key events, passes them to the hooks layer, where key events can be caught and additional actions triggered. Events that are not caught then arrive at the sluice, where event processing can be held. If held and then released, cached events are being re-inserted into dispatch. If not held, events are being processed and lookup of buttons occurs. In this system, releases are handled via hooks: once a button press is handled after passing through the sluice, a hook is registered that emits the correct key release upon receiving the corresponding event.

The issue is that when a key has already been pressed before processing was held and the key is then released, its release event is being emitted immediately because the release hook has already been registered, and its emit statement is therefore not being held up at the sluice.

My proposed solution seems like the least invasive way to approach this and to reuse most of the existing infrastructure. Instead of having an `emit` statement in the release hook, we are injecting the corresponding key event into dispatch, therefore allowing it to be blocked by the sluice, just as presses are blocked.

A potential problem with this is that a release event could pass through the hooks layer twice, so one key release could trigger multiple release hooks to be fired. My proposed solution is to flag key events specifically as already having been processed, thus safeguarding them from the hooks and any lookup in the button hashmap. This PR introduces a `WrappedKeyEvent` type that wraps a key event and additionally has a `passthrough` field that indicates whether an event should be handled by the hooks system and lookup up in the hashmap or not.

I'm pretty inexperienced with Haskell (just started re-learning it), so the code might be pretty bad. I would consider it a starting point to discuss this or other solutions to #256 and would love some comments to improve my coding style. In particular, one issue I had is that in some places, I wasn't able to use the (lenses type?) lookup with `(^.)` for the fields of the new type I introduced. I would love to know how to do that better.


jchtt:
Maybe another way to solve this issue would be to allow a hook to be registered as a sort of "master hook", holding processing of any other hooks. In that case, this would obviate the need for having a sluice: buttons like `tap-hold-next` could just register themselves as master hooks, taking a record of all events happening while they are active, and then taking care of injecting them into dispatch themselves.

jchtt:
I revisited this question and made another attempt at finding a cleaner solution that would work also with layers switches.

My current proposed solution introduces an additional hooks layer after the sluice in the pull chain and a WrappedEvent data type that can hold either KeyEvents or Uniques (hook tags from expired timers). Regular hooks are registered in this second hooks layer, while any hook that makes use of the `hold` function is registered in the first hook layer that I renamed `inputHooksPrio` (for priority). Moreover, timers are now taken out of the `injectTmr` in Dispatch, not in Hooks.

These changes mean that processing of any kind of event can now be paused, including key release events and timers. When paused, only the priority hooks layer will process these events, taking care of eventually re-opening the sluice.

@david-janssen @slotThe @thoelze1 @MuhammedZakir Let me know what you think!

MuhammedZakir:
I haven't looked into the KMonad's internals nor do I have necessary experience, so I can't comment on your code. But I will test it later; may take a few weeks though.

---

It looks like you accidentally modified `startup/run` file? https://github.com/kmonad/kmonad/pull/524/files#diff-fa2b73773005be60f40b7a2ec4ce791b241369aa8fee2b6bfec043145167acf7

jchtt:
> I haven't looked into the KMonad's internals nor do I have necessary experience, so I can't comment on your code. But I will test it later; may take a few weeks though.

Thanks, that would be great!

> It looks like you accidentally modified `startup/run` file? https://github.com/kmonad/kmonad/pull/524/files#diff-fa2b73773005be60f40b7a2ec4ce791b241369aa8fee2b6bfec043145167acf7

Yeah, I don't know what's going on there, it somehow automatically did this whenever I made a new branch. I suspect something weird is going on with my git on Windows.

meain:
Thanks a lot for working on this. I was able to test this out and seems to be working well for me. I'll keep using it and let you know if I run into any issues.

slotThe:
@jchtt sorry, life got in the way.  I will definitely try to get to it this week!

jchtt:
Thanks @slotThe, no rush! Not sure this is the best way to solve this problem, happy to think about other approaches as well.

slotThe (commented):
Just checking in to say that I haven't forgotten about this!  I've been running on this PR for a few days now and I quite like the behaviour (and haven't noticed any drawbacks so far).  

I would agree with your assessment that it's probably the least invasive solution with the current way that things work.  Having this two-hook solution is perhaps conceptually a bit "meh", but in this case I think perfection can wait until after `model-refactor` is not just a distant dream anymore :)

I plan to read the code in more details (hopefully soon...) but I think the overall idea is sound and from a glance I don't suppose will change much aside from some style considerations.  Thanks!

jchtt:
Thanks @slotThe. One style question I have is the following: if we keep the strategy as is, any hook function that calls `hold` should register a "priority" hook instead of a regular one. Is there any nice way of enforcing that?

MuhammedZakir:
I think we should document when to use `inputHooksPrio` over `inputHooks`, preferably with an example (use a currently defined button for that). This would help when defining new buttons.

slotThe:
> Thanks @slotThe. One style question I have is the following: if we keep the strategy as is, any hook function that calls `hold` should register a "priority" hook instead of a regular one. Is there any nice way of enforcing that?

Beyond making it a completely new type?  _Probably_ not.  A sleep-deprived, rambling, thought: we could perhaps make this work with phantom types, exporting not the data constructors themselves but constructor functions for them.  This needs some further deliberation but it's probably worth it

> I think we should document when to use `inputHooksPrio` over `inputHooks`, preferably with an example (use a currently defined button for that). This would help when defining new buttons.

Definitely yes!

david-janssen:
Just popping in to give a strategic answer: if you managed to identify a problem with the pullchain model and found a way to fix it, then I applaud that greatly. If we test things and they concretely fix a problem without introducing any new ones, I am all for it. I have 0 attachment to the direction the pullchain model is taken into, because I think that at its core, it's a very flawed solution written by someone who was still thinking imperatively.

I have loads of work to do in IO at the moment, so that's where I'll be focusing my efforts. If I do ever manage to wrap up that work, I'm turning my attention to a completely different way of handling the model. Until that time, any duct-tape on the pullchain is deeply appreciated :-)

src/KMonad/Model/Action.hs
```diff
@@ -93,8 +97,10 @@ makeClassy ''Trigger
 
 -- | ADT signalling where to install a hook
 data HookLocation
-  = InputHook  -- ^ Install the hook immediately after receiving a 'KeyEvent'
-  | OutputHook -- ^ Install the hook just before emitting a 'KeyEvent'
+  = InputHookPrio   -- ^ Install the hook immediately after receiving a 'KeyEvent'
+                    -- and before the sluice
+  | InputHook       -- ^ Installthe hook after the sluice
```
jokesper, line 102: I would have added a parmater to the `InputHook` to differentiate whether before or after the sluice.
This would make the `within` function more sensical as `within OutputHook` is a bit wierd.

src/KMonad/App/Main.hs
```diff
@@ -84,16 +84,19 @@ initAppEnv cfg = do
...
+  ihk <- Hs.mkHooks (Sl.pull slc) itr
 
   -- Initialize the button environments in the keymap
   phl <- Km.mkKeymap (cfg^.firstLayer) (cfg^.keymapCfg)
 
   -- Initialize output components
   otv <- lift . atomically $ newEmptyTMVar
-  ohk <- Hs.mkHooks . atomically . takeTMVar $ otv
+  otr <- atomically $ newEmptyTMVar
+  ohk <- Hs.mkHooks ((atomically . takeTMVar) otv >>= pure . (WrappedKeyEvent NoCatch)) otr
```
jokesper, line 99: When changing the parameter order of `mkHooks` as described above this too would be more readable\*:
```suggestion
  ohk <- Hs.mkHooks otr $ pure . WrappedKeyEvent NoCatch <=< atomically . takeTMVar $ otv
```
\* readablitiy is subjective so this is just my opinion.

src/KMonad/Model/Dispatch.hs
```diff
@@ -58,20 +60,21 @@ import qualified RIO.Text as T
...
+  , _rerunBuf :: TVar (Seq WrappedEvent)    -- ^ Buffer for rerunning events
+  , _injectTmr :: TMVar Unique -- ^ Used to signal timeouts
   }
 makeLenses ''Dispatch
 
 -- | Create a new 'Dispatch' environment
-mkDispatch' :: MonadUnliftIO m => m KeyEvent -> m Dispatch
-mkDispatch' s = withRunInIO $ \u -> do
+mkDispatch' :: MonadUnliftIO m => m KeyEvent -> TMVar Unique -> m Dispatch
+mkDispatch' s itr = withRunInIO $ \u -> do
```
jokesper, line 70: As mentioned above I would change the order of parameters like this and also for `mkDispatch`, `mkHooks`, ... (and maybe `Dispatch`, ...)
```suggestion
mkDispatch' :: MonadUnliftIO m => TMVar Unique -> m KeyEvent -> m Dispatch
mkDispatch' itr s = withRunInIO $ \u -> do
```

src/KMonad/Model/Dispatch.hs
```diff
@@ -82,7 +85,7 @@ mkDispatch = lift . mkDispatch'
 -- 1. The next item to be rerun
```
jokesper, line 85: I would add the timer in the comments of what is happening.

src/KMonad/App/Main.hs
```diff
@@ -84,16 +84,19 @@ initAppEnv cfg = do
   src <- using $ cfg^.keySourceDev
 
   -- Initialize the pull-chain components
-  dsp <- Dp.mkDispatch $ awaitKey src
-  ihk <- Hs.mkHooks    $ Dp.pull  dsp
-  slc <- Sl.mkSluice   $ Hs.pull  ihk
+  itr <- atomically $ newEmptyTMVar
```
jokesper, line 87: Apply hint
```suggestion
  itr <- newEmptyTMVarIO
```
This is a problem at a few points.
Even better rebase on `master` since their it's already fixed.

jokesper (commented):
This is what I would have personally changed. Keep in mind, I'm just a contributor I cannot actually merge this.
If I understand it correctly you also did a little refactoring where you touched the code
with which function handles timeout (missing understanding, not against it).
Nitpicking aside, nice work!

jokesper:
While trying to implement a fix for #955, I noticed that when we hold and catch an event via `InputHookPrio`, other input hooks can capture it when we stop holding it.
Is this intentional
`runHooks` runs every hook on an event.
But with holding we do the same, except that there may be a time difference in between.
We could simply not reregister events which have been caught.

Issue #525, opened 2022-04-18
Undocumented but useful "tap-macro-release" function.


mmarshall540:
EDIT: Originally, this was about my desire to create a software-only Vim-clutch using a footswitch with Kmonad. I've now located the function that makes this possible and have updated the title of the issue. The needed function exists already, but it's only documented in the source code. See my follow-up comment for further explanation.

Original post below:

Online guides describing the "[Vim-clutch](https://github.com/alevchuk/vim-clutch)" present it as a hardware project. But I thought it should be possible to do the exact same thing in software only, using Kmonad.

A Vim-clutch is a device for modal editing (where you have separate command and insert modes that you have to move between, depending on whether you're editing text or inserting text).

There are many cheap foot-pedals you can buy which act like a single-key keyboard. I have one such pedal and have confirmed that it's configurable with Kmonad, just as any keyboard would be.

But the idea with the Vim-clutch is that it sends one key when you push the pedal down (typically "i" for entering insert mode), and a different key when you release the pedal (typically "Escape" for going back to command mode). This allows one to hold the pedal when inserting text and then release it when done.

Reviewing the tutorial, I couldn't find a function for sending one key on press and a different one on release. I did notice the `cmd-button` macro, which can send a shell command on press and then a different shell command on release.

Is this already a feature that I overlooked?

mmarshall540:
I have just found the solution to this.

Thinking either that this feature must already exist or that I (without knowing any Haskell) might be able to cobble it together from existing code, I browsed the source code.

I already knew about the `tap-macro` function from the tutorial and reference. But that one doesn't work for this purpose, because it presses the last button before the trigger button's been released.

But when I looked through the source code, there was the similarly-named `tap-macro-release` function, which does **exactly** what I need. It presses the keys listed first, but waits to press the last key until the trigger key is released.

I now have a Vim-clutch built with software. No hardware mods required at all.

I've used GitHub's search mechanism repeatedly and can't find any mention of `tap-macro-release` in the documentation. But I think it's likely that others will want to use a footswitch this way.

PR #526, opened 2022-04-18
Document the "tap-macro-release" button.

mmarshall540:
Resolves: Undocumented but useful "tap-macro-release" function. #525 

This button was undocumented, but I was looking for the functionality it provides in order to make use of a footswitch as a "Vim-clutch". There's more explanation in the linked issue.

Here, I added some description of the button to the reference doc and also some explanation and an example in the tutorial.

I tried to make these fit with the style of the existing texts. Thank you!


slotThe:
Oh, indeed, this was added in #257 but we forgot to document it then.  Thank you for fixing that!

PR #533, opened 2022-05-03
NixOS module: Optionally make systemd group for multiple config services

rskew:
I wanted to switch off the service for my bluetooth keyboard when it's disconnected, but when the services are grouped they all shut down.

This change optionally stops the service group being created, allowing them to be started/stopped individually.

### DIFF
--- a/nix/module-base.nix
+++ b/nix/module-base.nix
@@ -40,6 +40,14 @@ with lib;
         The kmonad package.
       '';
     };
+
+    make-group = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Whether to make a service group when multiple configs are present.
+      '';
+    };
   };
 
   config = {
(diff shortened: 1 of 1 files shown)


nix/module-base.nix
```diff
@@ -56,7 +64,7 @@ with lib;
     systemd = with lib; with builtins;
       let
         # If only one config file is supplied, unify all kmonad units under a target
-        make-group = (length cfg.configfiles + length cfg.optionalconfigs) > 1;
+        make-group = cfg.make-group && (length cfg.configfiles + length cfg.optionalconfigs) > 1;
```
MuhammedZakir, line 67: I feel like the conditional here is wrong. The comment says "if only **one** config file ...", but the conditional says "`make-group` is `true` if there are more than one config file" [1]. Shouldn't it be `(<length of configs>) == 1`? Or, is the comment that's wrong?

[1] which results in below check being true if there are more than one config when it shouldn't?
```nix
if make-group
then { partOf = [ "kmonad.target" ]; }
else { inherit wantedBy; }
```

rskew: Good pickup! I'm pretty sure it's the comment that's wrong, as it doesn't make sense to make a group if there's only a single service.
I've updated the comment so it's congruent with what the code does (and used to do).

slotThe:
Looks reasonable to me; thanks!

Issue #252, opened 2021-05-16
[Windows] tap-hold-next-release emits both hold and tap button when held


MaxGyver83:
When I hold a `tap-hold-next-release` button for a while (I suppose it's the timeout of 500 ms I have set), I get the key emitted that results from both holding and tapping this key. But this only true for some of my home row modifiers.

Minimal example:
The `j` key is a modifier for `level4` (or _layer 4_) when held but a `t` when tapped. On layer 4, the `j` key is a `4`. When I hold the `j`, I get a `4`. (I expect that no key is emitted.)

```clojure
(defcfg
  ;; For Windows
  input  (low-level-hook)
  output (send-event-sink)
  fallthrough true
)

(defsrc j)
(deflayer level1 @t4)
(deflayer level4 4)

(defalias
  lv4 (layer-toggle level4)
  t4  (tap-hold-next-release 500 t @lv4)
)
```

[This is my full config](https://github.com/MaxGyver83/dotfiles/blob/master/kmonad/.config/kmonad/vou-win.kbd).
This bug occurs when I hold `a`, `f`, `j`, `;` or `c` (=`layer-toggle`) but not when I hold `s`, `d`, `k`, `l` (Shift or Ctrl) (all QWERTY names).

It happens both with version 0.4.1 and the latest commit. But only in Windows, not in Linux.

david-janssen:
Hmmm, that sounds like it might have something to do with how Windows and Linux represent key-repeat differently.

Linux always sends 1 Press events, then a bunch of Repeat events, and finally a Release event.
Windows sends a bunch of Press events, then a Release event.

Maybe we aren't correctly filtering out all of the presses in the Windows keyIO stuff?

Not going to tackle this issue right now, since I am in the process of redoing the keycode stuff anyways, but that means that I will be getting to the windows-key-io stuff soonish anyways. Then I will investigate this further.

gerhard-h:
Have the same Problem, a workaround could be to leave the button _ unmapped on level4 

 Windows  10
 kmonad.exe -V
0.4.1, commit 744c0445ac9beaece489cd6cdc981eeae45467a7

MaxGyver83:
This would work if I had only one level4 `tap-hold-next-release` key. I have two and both have a meaning in level 4. But I could create two layers: `level4-left-hand` and `level4-right-hand`. Then I could apply your workaround in both layers either for the left-hand or the right-hand key. I think I'll give it a try!

PR #541, opened 2022-05-18
Feature/win key repeat

david-janssen:
I think I have this working now. The way it's currently implemented, you may optionally provide 2 arguments to `send-event-sink` (it's optional whether you provide them, but when you provide them you must provide 2) that determine the `delay` and the `rate` of the key repeat.
`delay` - How many ms before a key starts repeating
`rate` - How many ms between each repeat event

I only have Windows running in a VirtualBox, so my testing setup is very limited. Let's try to get some people to try it and if nothing blows up we merge this in?

### DIFF
--- a/keymap/testing.kbd
+++ b/keymap/testing.kbd
@@ -1,0 +1,9 @@
+(defcfg
+  input (low-level-hook)
+  output (send-event-sink)
+)
+
+(defsrc q w e r t y lsft)
+
+(deflayer base
+  q w e r t y lsft)
--- a/keymap/testing.kbd
+++ b/keymap/testing.kbd
@@ -1,6 +1,6 @@
 (defcfg
   input (low-level-hook)
-  output (send-event-sink)
+  output (send-event-sink 1000 500)
 )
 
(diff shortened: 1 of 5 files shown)


slotThe:
Just pinging known Windows users: @6gk @whitbur @MaxGyver83

6gk:
ooo interesting
will check it out tomorrow!

MaxGyver83:
I have tested it:
It works! Awesome! I suppose it's the expected behaviour that it does not work for tap-hold-next-release keys.

(An undesired side effect of #252: When I press and hold the A key which is a tap-hold-next-release+layer-toggle layer 3 key, the according layer 3 key is being repeated.)

6gk:
Works really well at first look - thank you!for including an easy way to change the repeat rate, too.

6gk:
Works really well at first try - thank you!

The easy method of changing the repeat rate is much appreciated too.

david-janssen:
Very good, all automated tests are passing, all actual users seem pleased. Trigger pulled.

Issue #535, opened 2022-05-09
Problems with keycode 193 in Windows


fontanaricardo:
I'm using KMonad, with layout ABNT2, and every time that I press the key for `/` character, the prompt show's the error message `Encountered error in KeySource: Cannot translate from windows keycode: 193`.
After the message, KMonad stops to work.

My source configuration:

```
(defsrc
  esc     1    2    3    4    5    6    7    8    9    0    -    =    bspc
  tab     q    w    e    r    t    y    u    i    o    p    '    [
  caps    a    s    d    f    g    h    j    k    l    -    -    ]    ret
  lsft    \    z    x    c    v    b    n    m    ,    .    ;    /    rsft
  lctl    lmet lalt           spc                      ralt -    cmp  rctl
)
```

Using the precompiled version `kmonad-0.4.1-win.exe`.

br-lemes:
I released a binary for this fix: https://github.com/br-lemes/kmonad/releases/download/592c0fa/kmonad.exe

I've compiled it from the master branch with all fixes and new features since 0.4.1, but with all possible bugs as well.

br-lemes:
I released a binary for this fix: https://github.com/br-lemes/kmonad/releases/download/592c0fa/kmonad.exe

I compiled it from the master branch with all fixes and new features since 0.4.1, but with all possible bugs as well.

slotThe:
Closed via https://github.com/kmonad/kmonad/pull/546

renan-lopes-eso:
In the version 0.4.4 with the ABNT2 layout, pressing the physical key Fn causes the same problem:
"Encountered error in KeySource: Cannot translate from windows keycode: 237":

https://github.com/kmonad/kmonad/issues/1038

PR #546, opened 2022-05-25
Fix KeyRo for Windows (slash key for some keyboard layout such as ABNT2) #535

br-lemes:
Fix issue #535


slotThe:
Looks fine to me; thanks!

PR #549, opened 2022-05-27
Replace existing Nix code with a Nix flake

pjones:
Using the New Nix Flake
=======================

For Development
---------------

If you use [direnv](https://direnv.net/) then the included top-level
`.envrc` will load the development environment into your shell/editor
for you.

Otherwise, change into the `nix` directory and run `nix develop` and
you will be dropped into a shell with all of the Haskell tools you
expect.

Using the KMonad NixOS Module
-----------------------------

Create your own `flake.nix` that looks something like this:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/release-22.05";
    kmonad.url = "github:kmonad/kmonad?dir=nix";
  };

  outputs = { self, nixpkgs, kmonad }: {
    nixosConfigurations.my_host_name = nixpkgs.lib.nixosSystem {
      system = "x86_64-linux";
      modules = [
        kmonad.nixosModules.default
        ./path/to/configuration.nix
      ];
    };
  };
}
```

More information can be found here:

  <https://nixos.wiki/wiki/Flakes>

Building KMonad with Various Versions of GHC
--------------------------------------------

    $ cd nix
    $ nix build .#kmonad-ghc8107
    $ nix build .#kmonad-ghc902
    $ nix build .#kmonad-ghc922

Running the NixOS Virtual Machine Tests
---------------------------------------

    $ cd nix
    $ nix flake check

PR #550, opened 2022-05-28
Revert "Replace existing Nix code with a Nix flake"

david-janssen:
Reverts kmonad/kmonad#549


david-janssen:
D'oh, I should merge things without testing or before coffee

PR #551, opened 2022-05-28
Replace existing Nix code with a Nix flake

david-janssen:
Note
-----

This was originally #549, which through my administrative inadequacy got prematurely merged and reverted. This pull request is a duplication of the former, so that the issues therein can be addressed prior to merging.

Using the New Nix Flake
=======================

For Development
---------------

If you use [direnv](https://direnv.net/) then the included top-level
`.envrc` will load the development environment into your shell/editor
for you.

Otherwise, change into the `nix` directory and run `nix develop` and
you will be dropped into a shell with all of the Haskell tools you
expect.

Using the KMonad NixOS Module
-----------------------------

Create your own `flake.nix` that looks something like this:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/release-22.05";
    kmonad.url = "github:kmonad/kmonad?dir=nix";
  };

  outputs = { self, nixpkgs, kmonad }: {
    nixosConfigurations.my_host_name = nixpkgs.lib.nixosSystem {
      system = "x86_64-linux";
      modules = [
        kmonad.nixosModules.default
        ./path/to/configuration.nix
      ];
    };
  };
}
```

More information can be found here:

  <https://nixos.wiki/wiki/Flakes>

Building KMonad with Various Versions of GHC
--------------------------------------------

    $ cd nix
    $ nix build .#kmonad-ghc8107
    $ nix build .#kmonad-ghc902
    $ nix build .#kmonad-ghc922

Running the NixOS Virtual Machine Tests
---------------------------------------

    $ cd nix
    $ nix flake check


david-janssen:
I am running into errors when I try to execute `direnv allow` or `nix develop`. I'm not sure if these errors are a problem with the nix-code in this pull-request, or if I am not running the right nix-setup (e.g. `nix` version and `nix-channel` settings) on my own system.

Trying to enable direnv
```
$ direnv allow
direnv: loading ~/prj/kmonad/.envrc
direnv: using flake /home/david/prj/kmonad/nix
error: 'print-dev-env' is not a recognised command
Try '/nix/store/93p5wvnxngf7xi1xq5agrs61amsy0fp4-nix-2.3.16/bin/nix --help' for more information.
error: 'flake' is not a recognised command
Try '/nix/store/93p5wvnxngf7xi1xq5agrs61amsy0fp4-nix-2.3.16/bin/nix --help' for more information.
```

Trying to nix develop
```
(./nix)$ nix develop
error: flake 'git+file:///home/david/prj/kmonad?dir=nix' does not provide attribute 'devShells.x86_64-linux.devShell.x86_64-linux', 'packages.x86_64-linux.devShell.x86_64-linux', 'legacyPackages.x86_64-linux.devShell.x86_64-linux', 'devShell.x86_64-linux' or 'defaultPackage.x86_64-linux
```

pjones:
@david-janssen I think the issue is because your version of nix is 2.3.  I believe you'll need at least 2.4 for this.

david-janssen:
Yeah, I just started diving into the issue. My system-wide `nix` is 2.4, and wherever I ask `nix --version` that's what it tells me. I'm as we speak trying to upgrade my direnv, because the error does show a path to a `nix` of 2.3, so I'm guessing that's the one that my direnv has been installed with?

However, I am still very unsure about the second issue. Calling `nix develop` in the nix folder also doesn't work, and if I call `nix --version` in that subdirectory it reports 2.4, so I don't think fixing `direnv` will fix that issue.

david-janssen:
So yep, bumping up my nixos channel to `22.05` did the trick for the first issue. I'm guessing that `direnv` in `21.11` was using `nix` 2.3.

My second issue remains if I simply call `nix develop` and I discovered that `nix build` causes the same error. When I specify the ghc-version to use with the `.#kmonad-ghcxx-x` syntax you showed, both `build` and `develop` work. 

I'll leave this open till tomorrow, so if you have an idea of what could be weird about the unqualified `nix build` or `nix develop` commands, then maybe a quick fix would be cool. Otherwise I'll just merge, because it does indeed work as expected. Thank you very much for your contribution :-D

david-janssen:
So, I just rebased my local feature branch that I'm working on onto this branch, and everything just works. It's lovely. Thank you so much.

I think the only issue I had with `develop` is that the documentation said I should just `nix develop`, which doesn't work. But I can now drop in to whatever ghc version I want by using the `nix develop .#kmonad-xxxx` (or `.#default`) syntax, which is great.

I'll merge this first thing tomorrow if I hear nothing further.

pjones:
I think I figured it out.  I'm using Nix 2.8.1:

```
$ nix --version 
nix (Nix) 2.8.1
$ nix develop
> ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.0.2
```

In Nix 2.8 they removed flake attributes like `devShell` and replaced it with `devShells.default` which I use in the flake that's part of this PR.  Maybe Nix 2.4 can't figure that out?

pjones:
I just realized there's a bug in the dev shell.

```
$ nix develop .#kmonad-ghc8107
> ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.10.7
> haskell-language-server --version
haskell-language-server version: 1.7.0.0 (GHC: 9.0.2) (PATH: /nix/store/fm9zi3yl0cn6qr94h6w8cp74sidxasdr-haskell-language-server-1.7.0.0/bin/haskell-language-server)
```

It drops you into a shell that has the correct compiler but the wrong
shell tools.  I'll have a fix here in a bit.

pjones:
I pushed a commit that adds dev shells that match the various flake packages.  The upside is that all tooling is built with the same compiler.  The downside is that some of the tools no longer compile:

```
$ nix develop .#kmonad-ghc8107                      
error: builder for '/nix/store/dgfjbpx37w3z8qpy8rsavmj4y7ghx81v-cabal-fmt-0.1.5.1.drv' failed with exit code 1;
       last 10 log lines:
       >         ‘C.FieldGrammar PrettyParsec FieldDescrs’
       >    |
       > 75 | instance C.FieldGrammar PrettyParsec FieldDescrs where
       >    |                         ^^^^^^^^^^^^
```

An alternative would be to use the current compiler (9.0.2) for tools that don't care which compiler they use (e.g., cabal-fmt) and the matching compiler for tools that do (e.g., haskell-language-server).

david-janssen:
>  The downside is that some of the tools no longer compile

None of the dev-tools that I use personally seem to be affected. Which brings me to an open question on something that I am unsure about. This is the first coding project that I'm administrating, so: what's a good practice? Does one usually host a common suite of dev-tools in a repo? Is it better to let users pull in their own tools at their own leisure? Is there some common-sense wisdom about this topic, or do different projects do things in different ways and this is more a 'pick your poison' type situation? 

I'm happy to prune the dev-tools down to:
- cabal-install
- haskell-language-server
- hlint
- stack (just so I can make sure `stack build` works)

@slotThe: your input about the above question would also be appreciated.

slotThe:
> None of the dev-tools that I use personally seem to be affected. Which brings me to an open question on something that I am unsure about. This is the first coding project that I'm administrating, so: what's a good practice? Does one usually host a common suite of dev-tools in a repo? Is it better to let users pull in their own tools at their own leisure? Is there some common-sense wisdom about this topic, or do different projects do things in different ways and this is more a 'pick your poison' type situation?

I have to say that I'm quite inexperienced with nix and best practices.  I've never touched the xmonad nix flake at all, but I doubt we're even including HLS in there.  This

> I'm happy to prune the dev-tools down to:
>
>     * cabal-install
>     * haskell-language-server
>     * hlint
>     * stack (just so I can make sure `stack build` works)

sounds like a good set of tools to me! (to be honest, I don't think running kmonad with a different version of GHC will be something that a lot of users want to do anyways, so in the end it's probably not all that important)

david-janssen:
Very good. If I hear nothing by tomorrow, I will merge this branch, and then edit the dev-tools down to the ones proposed above (+ maybe stylish-haskell)

david-janssen:
Wahay!

PR #552, opened 2022-05-31
Feature/gestures

david-janssen:
I wanted to redo keycode representations,
- but its semantics leaked into the parser

I wanted to redo the parser
- but its semantics leaked into the invocation

I wanted to redo the invocation
- but its semantics leaked into the app-loop-configuration

I wanted to redo the app-loop-configuration
- but I was lacking a way to easily express compound keyboard actions.

This feature branch fixes that by introducing a small DSL for 'gestures', sequences of key events that are guaranteed to be valid (i.e. no pressing of already pressed keys, no releasing a key before it's been pressed, etc), plus a mini-language so that people can express all possible (I think..) gestures.

Also, I copied over some of the testing-framework from keycode-refactor into master and wrote a few tests for Gestures. 

This is basically ready to merge, and I was about to push it straight to master, but then I remembered that the better part of valour is discretion. So, any issues/critiques?

### DIFF
--- a/kmonad.cabal
+++ b/kmonad.cabal
@@ -39,6 +39,7 @@ library
   build-depends:
       base
     , cereal
+    -- , hspec
     , lens
     , megaparsec
     , mtl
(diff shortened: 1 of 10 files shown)


david-janssen:
Already discretion has payed off. Old LTS'es were missing an implicit import of 'fromEither'. Fixed it by making the export explicit. (I think)

PR #554, opened 2022-06-01
Starting to restructure configuration 

david-janssen:
This feature branch shows the ideas I'm exploring on how to make KMonad configuration more principled and less home-spun. 

The idea is to use Dhall as a programmable, extensible configuration language that we can pull straight into Haskell without having to write any parsers ourselves. Additionally, this pull request highlights how we could provide an interface for users to specify keyname-keycode relationships, while representing keycodes as Naturals. That way we will never encounter an untranslateable keycode again, and everyone can define their own keycode names.

There is a lot of work to be done, this pull-request only includes sketches of ideas that are not integrated into KMonad yet. However, I'll be working on the config format, and then move to replacing the `(defcfg..)` blocks and changing how KMonad gets invoked, then finally I'll end up replacing 'Keycode' with 'Natural', and hopefully most locale/OS/keycode-translatino issues will be a thing of the past.

I'm not sure we'll be able to keep it backward compatible, (actually, I'm quite sure we won't), but omelets, eggs, sumthin sumthin

### DIFF
--- a/src/K/Layers/Initial.hs
+++ b/src/K/Layers/Initial.hs
@@ -98,6 +98,6 @@ newtype Layer c = Layer { _layer :: [(c, But c)]}
   deriving (Eq, Show, Functor)
 
 data KeymapCfg = KeymapCfg
-  { _layers  :: Named (Layer Keycode)
-  , _aliases :: Named (But Keycode)
+  { _layers  :: Labeled (Layer Keycode)
+  , _aliases :: Labeled (But Keycode)
   }


david-janssen:
> 1. If a keycode for a keyname that is used in `*.kbd` is not found for the the current OS, rather than exiting, log an error and use `XX` (nop) for that keyname.
>    
>    * A flag can be added to toggle this - i.e. whether to exit or continue running after logging an error.
>    * This is useful for cross-platform config. It's really a pain to debug the correct keycode in different OSes, especially if you only use those keys in specific OSes (e.g. Apple-spefic keys such as `launchpad`). So an escape hatch **really** helps.
>    * An optional shell command can be given which will be called with an error message. Useful when you are running KMonad in background and don't want to miss any error msgs, in case, it's a genuine error. Example: for shell command: `notify-send "KMonad Error"`, KMonad will execute `notify-send "KMonad Error" '<error msg>'`

So I think this would be a good fix for the current problem, but the actual intention is to set up a system that simply doesn't require this fix. The idea is based around the following:
1. Mac, Win, Lin keycodes can all be represented as Natural numbers. Win and Lin simply by definition, and Mac keycodes are 2 Word32's, so we can jam them together and represent them as a Word64. By taking 'Natural' as the base type, *any* countable set of keycodes can be represented, so if you can define a 'Keycode -> Natural' for your OS, then we can represent all possible 'Keycodes'.
2. There are 2 ways we are currently using keycodes in KMonad
  - For the parser, so when people say `(defsrc a ...)` we need to know what `a` is pointing at
  - For certain buttons, like `shifted` buttons, compose sequences, etc. E.g. we are using hardcoded `KeyShift` in the source-code.

So what if, instead: 
- We provide people with a way to give a 'Name -> Keycode' mapping (where Keycode is just an alias for Natural). 
- We provide people with a way to give a 'Name -> Gesture' mapping (where Gesture is some combination of output buttons, like `a` or `S-a` or `S-[a b c]`.

Then we can:
- get rid of *all* internal names for keycodes
- get rid of *all* internal names for sequences (the shifted names, the compose sequences, the hardcoded `C-` means control-around style buttons

For parsing the keymap we rely on the provided Locale (the 'Name -> Keycode' and 'Name -> Gesture' maps). If during this process we encounter an unrecognized name, that's an error, but it's *parse*-error, not a run-time error. If during runtime we encounter a keycode without a name that's not an issue. We won't have anything bound to it, because we can't refer to it in the config, so if fallthrough is True we just reemit, if fallthrough is False it just gets blocked. ezpz.

If a user encounters a keycode that we haven't provided a name for in our default configs it becomes very easy to add them to their KMonad through the dhall configuration (so no recompilation required). Something like this:

```dhall
let L = ./locale/linux/enUS.dhall -- To import a map for linux enUS
```

Then is becomes really easy to add buttons, so if you have a button on your keyboard that you can't bind to, just look at the debug-logging output of kmonad to see that the natural keycode coming from that button is 123 or 321 and add it to your keycodes like this:
```dhall
let mycodes = L.keycodes // { weird = 123, red_button = 321 }
```

And then it also it becomes not-our-problem on how to deal with locales. Because users can just overwrite the keycode bindings in their config, something like this:
```dhall
let deDE = L.keycodes // { z = 35, y = 67 } -- Note, the numbers are made up
```

So then we have a system that can represent any keycode, can discover and report those codes to the user, and provides a system to bind names to those codes which can then be used in the config. Then we should have:
- No more unrecognized keycode crashes
- No more platform-translation issues
- No more giant modules of keycode bindings to maintain
- No more locale issues
- A much smaller parsing language to maintain

MuhammedZakir:
I completely agree with you and I am in favor of this. However, I am still confused on this part:

> For parsing the keymap we rely on the provided Locale (the 'Name -> Keycode' and 'Name -> Gesture' maps). If during this process we encounter an unrecognized name, that's an error, but it's parse-error, not a run-time error.

What happens when we get such parse-error? To make it clear, consider this situation:
1. I have a config that I use in 3 OSes.
 ```clojure
(defsrc  a  b  c  foo)

;; shift each letter by 1 pos
(deflayer  foo  a  b  c)
```

2. My keycode config:
    - `macOS.dhall`: `{ ... , foo = 1234 }`
    - **No keycode defined for `foo`** in `linux.dhall` and `windows.dhall`.

3. When I press `a` in macOS, KMonad will send `foo`'s keycode, `1234`. But what happens when I press `a` in Windows or Linux? As there is no key `foo` defined for these two OSes.

**Comment by david-janssen** [comment]
> 1. I have a config that I use in 3 OSes.
> 
> ```clojure
> (defsrc  a  b  c  foo)
> 
> ;; shift each letter by 1 pos
> (deflayer  foo  a  b  c)
> ```
> 
> 2. My keycode config:
>    
>    * `macOS.dhall`: `{ ... , foo = 1234 }`
>    * **No keycode defined for `foo`** in `linux.dhall` and `windows.dhall`.
> 3. When I press `a` in macOS, KMonad will send `foo`'s keycode, `1234`. But what happens when I press `a` in Windows or Linux? As there is no key `foo` defined for these two OSes.

Ahhhh, that's an interesting use-case I hadn't considered. If `foo` were undefined in linux or windows, then KMonad should error out before even attempting key-IO. Basically, the moment it reads the `kbd`-file and encounters a keyname or gesturename that isn't defined in its locale, I think it's proper for KMonad to report this and exit.

However, there is also a very easy, practical fix. You could just define `foo` as `999999999999` in the linux and windows dhall configurations, that way we can parse them, but they'll never occur. The other option would be to add a lot of annoying logic to the keymap parsing, which I really want to avoid.

Oh.. but wait.. then if you press `a` you'd be emitting `999999999` to the OS, which would be an issue.

Then I guess we *could* add a bunch of annoying logic to the parser, but who really wants to write and maintain that?

What if instead you bound them to some innocuous code? 

Basically, you could run KMonad, and push that evil button. If it registers an event with a keycode, then you can capture it and actually make that button useful. If the button *doesn't even register* then you want to bind a code that  you'll never send to KMonad, and that code should also be a code that has no effect when you accidentally emit it to your OS.

Maybe something like `KeyCoffee` from the LinuxHeaders. There are a bunch of weird keycodes that don't get used by the OS.

I think this practical fix is probably nicer, because it keeps the semantics and the parser clean. I.e. Keynames *must* exist, otherwise your configuration is not well-defined. Then make the user responsible for finding a definition that doesn't blow up their computer. At least that way we can ensure that KMonad will not be the thing that is blowing up.

**Comment by david-janssen** [comment]
> How about exposing XX to keycode config? XX is synonymous to NOP. 

If I understand correctly you mean no-op with `NOP`, but keycodes are not operations, they are only values. `XX` is not a value in the classic sense, `XX` only makes sense in the context of talking about things-kmonad-should-do, not in the context of making sense of a  an input signal of alternating states labeled with keycode-values. What does it mean to press a `NOP`? What does it mean to release a `NOP`? I don't understand how this would work.

> we can use 0? We can consider 0 as a special case.

0 comes with semantic load, at least in linux (input-event-codes.h: `#define EV_SYN  0x00`). Additionally, the problem with hard-coding a special case is that now we (the kmonad maintainers) become responsible for properly encoding all of the special-cases in KMonad, and dealing with all the little wrinkles that come with those cases. By providing the ability to bind arbitrary names to keycodes, but demanding that names must point to valid codes, the core of kmonad stays as simple and clear as possible, while moving all the edge-case logic to the users. 

The case you are describing: I want to use the same config file for 3 different OSes and I want KMonad to cleverly deal with keys that are OS-specific can be solved by either binding these weird keys to innocuous (but real) keys in your configuration, or by maintaining 3 different keymaps. I understand if neither of those options spark joy, but I hope you see that tracking a lot of edge-case logic inside KMonad sourcecode does not spark joy for me either :).

Of course, if I am misunderstanding you, then I apologize.

An alternative solution to the problem that I am considering, but that is also not within reach at the moment, is defining context-tags in KMonad. Something along the lines of buttons that decide what to do based on some value or variable that has been set in some dictionarly somewhere. Then you could have a button that does different things depending on if a `linux` tag is present, or an `osx` tag etc.  But again: faaaar down the line.

**Comment by MuhammedZakir** [comment]
> What does it mean to press a `NOP`? What does it mean to release a `NOP`? I don't understand how this would work.

I only did a quick look at KMonad's source and so I only have a vague understanding of it's architecture. So please bear with me. :-)

Note: when I mentioned "define key", you can consider it as "assigning anything other than `XX` for that key".

Coneceptually, if a key is assigned `XX`, then it is similiar to that key not existing in the captured keyboard. For instance, if I define `fn` in macOS, but assing `XX` in Linux, then KMonad will *never* get an `fn` event in Linux; if I actually press `fn` in Linux, then what KMonad gets is an *unknown event* and it will be handled just like other unknown events. This means that for `XX` keycode, we do not need to care about it in input-hadling-layer.

So what we only need to care about is: what will be the result of pressing `a` for below config in Linux if key `fn` is `XX`?

```clojure
(defsrc
  fn  a)

(deflayer main
  a  fn)
```

Before getting into that, just to make sure we're on the same page: `a` in layer `main` will **never* be triggered as KMonad will **never** get a `fn` key event.

Presuming KMonad checks whether a key encountered in the config is defined or not while/after parsing, then (in psudeo code):
```hs
import qualified Data.Set as Set (insert)

-- snip --

-- `xxKeys` is a set of keys whose keycode is "XX".
--
-- In above config, parsedKeyName == "a" or "fn".
isValidKey :: Set Keyname -> Keyname -> Either Keyname Keyname
isValidKey xxKeys parsedKeyName = 
  case lookupKeycode parsedKeyName of
    Just keycode -> do
      let xxKeys =
        if keycode == KeyXX
        then Set.insert parsedKeyName xxKeys
        else xxKeys
      Right xxKeys
    Nothing -> Left xxKeys
```

Once we have `xxKeys`, we can use it to essentially treat the above config as:
```clojure
(defsrc
  fn  a
)
(deflayer main
  a  XX
)
```

The same for aliases:
```clojure
(deflayer main
  a  (tap-hold fn a)
)
```
will become
```clojure
(deflayer main
  a (tap-hold XX a)
```

How to do this? I am not familiar with the codebase to give a good answer, but I guess when building "button"s? [1]

[1] Before `Gesture` feature, it would be here (I think): https://github.com/kmonad/kmonad/blob/31fec998b84451a5da3bfe4225869516e1611eee/src/KMonad/Args/Joiner.hs#L323-L324

david-janssen:
>  So please bear with me. :-)

Absolutely :-). I think from reading your analysis of how things work, there are a few misunderstandings that I should clear up.

> Note: when I mentioned "define key", you can consider it as "assigning anything other than `XX` for that key".

I think the misunderstanding we have starts here. KMonad doesn't actually 'assign buttons to keys' in a literal sense of that description. Instead, during parsing, it constructs a set of dictionaries (structured as a stack of layers) mapping keycodes to actions. Then, while it's running, when an event arrives, we check its keycode against our layerstack, and depending on the stackstate, we either find some mapping in some dictionary that is willing to 'capture' that event and do something, or none of the currently active layers want to match, and the key-event 'falls through' unmatched.

Rereading the above paragraph: I think that my explanation is still a bit unclear, so let's illustrate:

```lisp
(defsrc a b c)

(deflayer 1 p q  _)  ;; {a: emit-p, b: emit-q}
(deflayer 2 r _  _)  ;; {a: emit-r}
(deflayer 3 s XX XX) ;; {a: emit-s, b: blocked, c: blocked}
```

So, the code above defines a 3-key source layer, and then 3 different layers, and behind each layer, in pseudocode, is the kind of dictionary that would be generated in the model. Now:
- If the layerstack was `[2, 1]` (where left means first, so `2` above `1`) 
  - Then let's say an `a` arrives from the input keyboard: We check `2` first, it contains a mapping from `a` to `emit-r`, so we consider the event captured, do an `emit-r`, and continue waiting for input.
  - Instead let's say `b` arrives: We check `2` first, it doesn't contain any action for `b`, so we check the next layer. `1` does have a mapping for `b`, namely to `emit-q`, so we consider the event captured, do an `emit-q`, and get on with it.
  -  Instead let's say a `c` arrives: We check first `2`, then `1`, and then we reach the bottom of the stack without ever matching a `c`. This event is now considered uncaught, so we check the `fallthrough` setting. If `True`, then we emit the event unchanged to the OS, if `False` we just ignore the event. Either way, we are done now and start waiting for the next event.

- Now let's say the layerstack was `[2, 3, 1]`
  - Then the `a` happens just as above: we match in layer `2` and do an `emit-r`
  - `b`, however, doesn't get matched in `2`, so we go to layer `3` where we *do* match, `b` matches to `blocked` in layer `3`, so we consider the event caught, do nothing, and get on with it.
  - Likewise with `c`.

As can be seen from the above description, if any button-type is a `NOP`, it's actually `_` (transparent). We filter all transparent buttons out of our keymaps, since the default behavior (don't match and check the next layer) is what transparent is supposed to do. The `XX` instruction actually inserts a `pure ()` (Haskell for: do nothing) as an action into a dictionary, so that the dictionary will match, and we will capture the key. 

> Coneceptually, if a key is assigned `XX`, then it is similiar to that key not existing in the captured keyboard. 

So, I think from the description of the computation I gave above, it becomes clear how "a key is assigned" becomes very semantically vague. If a key is assigned `XX` like in layer `3` above, then we have a dictionary, with that keycode in it, waiting to match that keycode against an event and then do nothing.


> For instance, if I define `fn` in macOS, but assigned `XX` in Linux, then KMonad will **never** get an `fn` event in Linux;

No, if you define an `fn` key in Mac and assign `XX` to it in Linux, the layers are there trying to match `fn` to events. We *get* the events, we just explicitly do nothing when we get them. 

I've ignored the rest of the text before your example, because a lot of it comes from a misunderstanding that I hope to have clarified with my above explanation. Let's move on to your concrete example:

> ```clojure
> (defsrc
>   fn  a)
> 
> (deflayer main
>   a  fn)
> ```
>
> Before getting into that, just to make sure we're on the same page: `a` in layer `main` will **never** be triggered as KMonad will **never** get a `fn` key event.

Absolutely, we're on the same page :-), **iff** pushing the actual `fn` key on your keyboard does not generate an event in `evtest`. However, if pushing `fn` *does* generate an event, but just with a really weird code, then we could capture that code and execute actions, it's just that currently our cross-platform keycode support isn't ready, so theoretically possible, if not realized as of yet.

> ```haskell
> <snip>
>   case lookupKeycode parsedKeyName of
>     Just keycode -> do
>       let xxKeys =
>         if keycode == KeyXX
>         then Set.insert parsedKeyName xxKeys
>         else xxKeys
>       Right xxKeys
>     Nothing -> Left xxKeys
> ```

Forgive me if I'm belabouring the point, it is impossible to tell when writing textual responses when your reader has grokked your point :-). So, this code is simply impossible, and it's most clearly illustrated by looking at the following line:

> if keycode == KeyXX

There is no `KeyXX`

If instead you wanted to look at (very-pseudo-code):

> if (parse-result-of `a` in `defsrc`) == (parse-result-of `XX` in `deflayer`)

Then you would get a type error, because the result of parsing an `a` in a `defsrc` is a `Keycode`, and the result of parsing an `XX` in a `deflayer` is a `Button` (after a few in-between steps).

There is another wrinkle that could be contributing to your confusion.
```
(defsrc a)   ;; parsing a generates a Keycode
(deflayer 1 a) ;; parsing a generate a Button
```
Since all keycodes have a 'simple' action associated with them (simply emit them as a keycode, i.e. featureless, normal button) we automatically create a button for each 'Keycode' that has the same name as the keycode, that will function as a normal button that emits that keycode.

Note that there *is* a button `XX`, the do-nothing-button, but there is no `KeycodeXX` because... well.. that key doesn't exist. It's an entirely different beast.

> Once we have `xxKeys`, we can use it to essentially treat the above config as:
> 
> ```clojure
> (defsrc
>   fn  a
> )
> (deflayer main
>   a  XX
> )
> ```
> 
> The same for aliases:
> 
> ```clojure
> (deflayer main
>   a  (tap-hold fn a)
> )
> ```
> 
> will become
> 
> ```clojure
> (deflayer main
>   a (tap-hold XX a)
> ```
> 
> How to do this? I am not familiar with the codebase to give a good answer, but I guess when transforming parsed tokens to "button"s? [1]

So, I'm still not entirely sure what you want to do: do you think you could explain to me *what* you want without expressing it in terms of how it should work? There are certainly many things possible, and some of the stuff you describe above could be hacked into KMonad.  We could change layers from `Map Keycode Action` to `Map (Maybe Keycode) Action`, to add in a special keycode `Nothing` that represents codes that don't occur on this OS, but that would muddle the semantics tremendously, and force every single line of code that deals with event-matching to deal additionally with the option of it being `Nothing`. This has computational and maintenance overheads (computational, neglible, maintenance and ease of reasoning: tremendous).

I hope this clarifies the issue. I really appreciate you wanting to figure out how to get at the internals of KMonad. Please feel free to keep asking for clarification and features :D

MuhammedZakir:
Thanks for the clarification and bearing with me! :-)

---

_What I want to do with all these_:

Finding keycodes for OS-specific keys for different OSes is really troublesome and I am hoping to get some kind of escape-hatch to deal with it.

_What I hope will solve this problem_:

1. Some way to "mark" a key in the keycode-defintion file (`.dhall`).
2. These "marked keys" will be treated/translated as/to `blocked` (`pure ()`).

---

I have one more confusion: what happens when KMonad gets "unknown events"? For example, KMonad gets a press event of `100` and in the keycode-definition file, there is no key with keycode `100`. What will KMonad do?

david-janssen:
> Finding keycodes for OS-specific keys for different OSes is really troublesome and I am hoping to get some kind of escape-hatch to deal with (useful for cross-platform config).

That sounds like a very achievable goal and it's something I'm working towards. 

The reason why this works so poorly at the moment is because of how we have defined keycodes: essentially as a giant Enum of codes that we have assigned names to in the source code of KMonad. This was all based on Linux, so then we had to write a lot of translation code between the OSes. But suddenly all the edge-cases become huge issues. That's why the push to change keycode to `Natural`.

Do you have a specific key right now that breaks `kmonad` under Linux? Have you tried running `evtest` on the keyboard and pushing that specific button to see what gets printed out? Because if it's a keyevent, and it has a code, then it should be theoretically catchable. The issue being, it might not be defined in our giant Enum. So now, when the keycode arrives and we try to convert it we discover we can't, and KMonad crashes.

The idea of assigning names to keycodes is to get rid of the Enum altogether, represent keycodes as an infinitely large type, and then we don't need to convert at all anymore. Which means we wouldn't need an escape-hatch, because we'd have demolished the prison.

To make this easier one of the first things  I'm hoping to add after this branch gets merged is a run-mode for `kmonad` where you can just press keys, and `kmonad` tells you the numerical code, and if we have any names bound to that numerical code. If you then have a weird `fn` button and you push it, and `kmonad` says: `code: 42424242, names: <none>`, then you can bind a name to that number in your locale, let's say `foo`, then you can now refer to that keycode as `foo` in your configuration.

However, at the moment that's just a dream. I'm getting back to cfg-specification code :).

Cheerio!

MuhammedZakir:
> Do you have a specific key right now that breaks `kmonad` under Linux? Have you tried running `evtest` on the keyboard and pushing that specific button to see what gets printed out? Because if it's a keyevent, and it has a code, then it should be theoretically catchable. 

This is the problem I am trying to avoid -- _finding the keycode_. Take the `fn` key, `launchpad` key, or any other macOS-specific keys. I **only** want those in macOS, but neither in Windows nor in Linux. Simply because only the Macbook's internal keyboard has those. So if I want find the keycode of those keys in Windows or Linux, then I will have to spin a VM inside my Mac. The same situation applies if I have to find keycodes of Windows/Linux-specific keys [1] - have to spin 2 VMs for other OSes in each OS. You can imagine how troublesome this is especially considering you're doing all this to _just get KMonad running_. FWIW, what you said works if I have external keyboards that have these keys, but generally, these keyboards only have keys that are found in all major OSes.

tl;dr -- the solution works mostly, but is really troublesome and even infeasible, especially for a cross-platform config.

[1] It may not be "specific", but the situation applies in that it's hard to find the corresponding keycodes in other OSes.

david-janssen:
> So if I want find the keycode of those keys in Windows or Linux, then I will have to spin a VM inside my Mac.

2 things:

1. I'm a bit confused, if you'd have to go to a lot of trouble to figure out what the keycode would be for linux or mac, then why is it important your config be cross-platform stable? I.e. if you're so far away from running KMonad on Linux, why would your config need to be Linux compatible?

2. I'm hoping that the `evtest` mode for KMonad should make it very easy to discover what keycodes belong to what physical keys for the machine you're running KMonad on. That way if you do ever have to run on Linux, for example, you could fire up evtest mode and discover what numbers go with what keys on that particular setup in just a few seconds.

MuhammedZakir:
Consider this:
- A Macbook, 1 Windows computer and 1 Linux computer.
- I want to use KMonad on all of them.
- In the config, all the keys will be "common" keys, except in a layer, say `media`, that will contain media keys like `brightness up/down` and also OS-specific keys like `launchpad`.
- This means that, if I want to use this config in all 3 OSes, I will have to find keycodes of OS-specific keys like `launchpad` in other OSes.

This problem can be solved by either:
1. Find keycodes of OS-specific keys. For instance, for `launchpad` of macOS, find its keycodes in Windows and Linux.
    - Ths is what you're suggesting. But, if I want to do that, then I must have an external keyboard with those keys, so I can plug it into different OSes, or have to use VMs - in the above example, Linux and Windows VMs in macOS; which obviously, is not something I would want to do just to get KMonad running.
    - The problem becomes worse if there is a Linux- or Windows-specific key.

2. Use different configs for each OS; which I don't want to do unless there is no other choice.

3. Remove OS-spefic keys; which is also something I don't want to do unless there is no other choice.

4. Have a way to say to KMonad: "I know there is no valid keycode for this key; this is intentional. Treat/translate this as/to `KBlock` (`blocked` / `pure ()`) instead of `KEmit`".
    - https://github.com/kmonad/kmonad/pull/554#issuecomment-1146736520

MuhammedZakir:
Btw, what I am asking is not a blocker to this PR. Rather, it is something that's to be implemented on top of this PR. So if you have finished working on this, feel free to merge this.

david-janssen:
>  So if you have finished working on this, feel free to merge this.

Hehe, if only :)

In regards to your post explaining the 3-os computer setup you're running, I see what you are saying. I think that since we should no longer be translating keycodes (when this refactor is done) KMonad shouldn't be crashing anymore even when running into unknown keys. 

Let's stop theorycrafting for a bit and I'll just focus on getting it done. Then when we have a setup where we can actually try solutions, we can see if things need adjusting, or if the available tools are sufficient for a satisfactory solution.

At my current pace this won't be anytime soon anyways.

PR #555, opened 2022-06-02
NixOS module generates defcfg block only when requested

pjones:
This change updates the new NixOS module so that it will only generate
the `defcfg` block when requested.  The default behavior is to not
generate the `defcfg` block and therefore one must exist in your
KMonad configuration.

New option: `defcfg.enable` to enable `defcfg` generation.

Breaking change: options related to the `defcfg` block have been moved
under the `defcfg` attribute set.

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -33,29 +33,36 @@ let
         '';
       };
 
-      compose = {
-        key = lib.mkOption {
-          type = lib.types.nullOr lib.types.str;
-          default = "ralt";
-          description = "The (optional) compose key to use.";
+      defcfg = {
+        enable = lib.mkEnableOption ''
+          Automatically generate the defcfg block.
+
+          When this is option is set to true the config option for
+          this keyboard should not include a defcfg block.
+        '';
+
+        compose = {
+          key = lib.mkOption {
+            type = lib.types.nullOr lib.types.str;
+            default = "ralt";
+            description = "The (optional) compose key to use.";
+          };
+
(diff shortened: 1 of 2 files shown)


david-janssen:
Hey pjones, 

Thanks for the contribution. I am happy to merge this, but I want to quickly get your and @slotThe 's views on how best to deal with:
>Breaking change: options related to the defcfg block have been moved under the defcfg attribute set.

Ideally we would not break the user's experience with KMonad too often, and when we do it we would like it to be as gentle and well-motivated as possible. If I understand correctly, the nixos module was broken anyways before you started improving it, right? So then it would very much fall in the category of under-active-development, and I guess that breaking configuration is simply an affliction early adopters choose to endure.

I think I'm just unsure about annoying people, since I also intend to annoy them soon with a breaking change myself.

pjones:
@david-janssen Since the new NixOS module breaks backwards compatibility with the previous version, and since we haven't released the new module yet, I think it's okay to still change it.

I think it's understood that if you are using an unreleased version directly out of git you expect things to break from time to time.

david-janssen:
> using an unreleased version directly out of git you expect things to break from time to time

I shall try to keep that in mind and let go of my worries

slotThe:
> > using an unreleased version directly out of git you expect things to break from time to time
>
> I shall try to keep that in mind and let go of my worries

Yeah, I think one can reasonably expect HEAD to break from time to time.  It's one of the unexpected upshots of not having made a release in a while :)

PR #557, opened 2022-06-04
Bump compiler version in the Dockerfile

lierdakil:
Stack.yaml was updated to lts-18.27 resolver, which uses GHC 8.10.7;
the Dockerfile wasn't updated accordingly, though.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM lierdakil/alpine-haskell:8.10.4
+FROM lierdakil/alpine-haskell:8.10.7
 
 WORKDIR /usr/src/kmonad/
 RUN apk --no-cache add git


david-janssen:
Aaahhhh very good. I never use `docker`, so it didn't occur to me that this was necessary. Thank you for the fix! I'll try and remember in the future, but no guarantees ;)

lierdakil:
Huh. That broke the build :unamused: I'll see where I messed up shortly.

lierdakil:
Ah, I see, the tag was using ghc-8.10.7-integer-simple build (since that's what ghcup was providing at the time), and this project kind of depends on integer-gmp. I've updated the docker tag to use integer-gmp build, I believe no further action here is necessary, future builds should "just work".

david-janssen:
>  I've updated the docker tag to use integer-gmp build

Wait, do you need me to merge this? Does `master` docker work right now?

lierdakil:
> Wait, do you need me to merge this?

No, I force-updated the tag on docker-hub.

> Does master docker work right now?

Well, it should, at least it [works](https://github.com/lierdakil/kmonad/runs/6748433469?check_suite_focus=true) on [my fork](https://github.com/lierdakil/kmonad/commit/f9df30991c1adfdcf4e4ef1623b9e716704e40eb) (which is even with `master`). But you can always check by restarting the docker job on CI https://github.com/kmonad/kmonad/runs/6743668807 :smile:

david-janssen:
> No, I force-updated the tag on docker-hub.

I have no idea what most of those words mean, but I'm very happy to take your word for it :) Thanks!

PR #559, opened 2022-06-06
nixos-module: use submodule name as keyboard name by default

fufexan:
This will make using
```nix
keyboards = {
  internal-keyboard = {
    name = "internal-keyboard"; # optional now
    ...
  };
};
```
optional. Tested on my configuration.

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -9,6 +9,7 @@ let
       name = lib.mkOption {
         type = lib.types.str;
         example = "laptop-internal";
+        default = name;
         description = "Keyboard name.";
       };
 


david-janssen:
This looks fine to me, let's quickly check with @pjones, who is much more knowledgeable about the nix stuff. If we hear nothing from him in a day or two I'll just merge this.

pjones:
I don't believe this is necessary due to:

```nix
{
  keyboard = { name, ... }: {
    options = {
      # ...
    };
    
    config = {
      name = lib.mkDefault name; # This sets the default.
    };
  };
}  
```

If you look at the NixOS tests you'll see that the `name` attribute is never explicitly set.

fufexan:
Oh, my bad. I misread your message. It seems you are correct. Therefore I will close the PR.

Issue #79, opened 2020-09-13
Add .XCompose file parsing for Linux


david-janssen:
As described here, this might be a better way to deal with compose-key sequences:

https://github.com/david-janssen/kmonad/issues/77#issuecomment-691679914

PR #575, opened 2022-06-25
A sketch of how general Unicode could be handled on Linux.

leftaroundabout:
This uses the fact that at least on X11, characters can be specified by ctrl+shift+U followed by the hexadecimal representation of the Unicode code point.

The implementation is a big hack, but could be generalised to something that allows specifying different way to deal with Unicode in the `defcfg`.

IMO something general like this is much preferrable to the current detour via Compose key seqs. After all, that's basically using another configurable keyboard remapper to make up for the shortcomings of this configurable keyboard mapper...
Also, better Unicode support seems to be much requested in the issues.

### DIFF
--- a/src/KMonad/Args/Joiner.hs
+++ b/src/KMonad/Args/Joiner.hs
@@ -32,6 +32,8 @@ import KMonad.Model.Action
 import KMonad.Model.Button
 import KMonad.Keyboard
 import KMonad.Keyboard.IO
+import KMonad.Parsing (runParser, some)
+import KMonad.Args.Parser (buttonP)
 
 #ifdef linux_HOST_OS
 import KMonad.Keyboard.IO.Linux.DeviceSource
@@ -49,6 +51,7 @@ import KMonad.Keyboard.IO.Mac.KextSink
 #endif
 
 import Control.Monad.Except
+import Numeric (showHex)
 
 import RIO.List (headMaybe, intersperse, uncons)
 import RIO.Partial (fromJust)
@@ -364,6 +367,12 @@ joinButton ns als =
     KComposeSeq bs     -> do csd <- getCmpSeqDelay
                              c   <- view cmpKey
                              jst $ tapMacro . (c:) <$> isps bs csd
+    KUnicodeChar cb    -> do csd <- getCmpSeqDelay
+                             -- c   <- view unicodeKey
+                             let uniInputKeys = T.pack
+                                  $ "C-S-u " ++ intersperse ' ' (showHex (fromEnum cb) "") ++ " spc"
+                             case runParser (some buttonP) "" uniInputKeys of
+                               Right bs -> jst $ tapMacro <$> isps bs csd
     KTapMacro bs mbD   -> jst $ tapMacro           <$> isps bs mbD
     KTapMacroRelease bs mbD ->
       jst $ tapMacroRelease           <$> isps bs mbD
--- a/src/KMonad/Args/Parser.hs
+++ b/src/KMonad/Args/Parser.hs
@@ -23,6 +23,7 @@ module KMonad.Args.Parser
   -- * Building Parsers
   , symbol
   , numP
+  , buttonP
 
   -- * Parsers for Tokens and Buttons
   , otokens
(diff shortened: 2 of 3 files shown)


leftaroundabout:
For me, this works on Ubuntu 18.04 with configurations like e.g.
```
(deflayer Greek
  esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12        ssrq slck pause
  grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc  ins  home pgup  nlck kp/  kp*  kp-
  tab  Ψ    Ω    Ε    Ρ    Τ    Υ    Υ    Ι    Ο    Π    [    ]    ret   del  end  pgdn  kp7  kp8  kp9  kp+
  esc  Α    Σ    Δ    Φ    Γ    Θ    Θ    Κ    Λ    ;    '    \                          kp4  kp5  kp6
  lsft home Ζ    Χ    Ξ    Φ    Β    Ν    Μ    ,    .    /    rsft            up         kp1  kp2  kp3  kprt
  lctl lmet lalt           spc                 ralt rmet cmp  rctl       left down rght  kp0  kp.
)

(deflayer greek
  esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12        ssrq slck pause
  grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc  ins  home pgup  nlck kp/  kp*  kp-
  tab  ψ    ω    ε    ρ    τ    υ    υ    ι    ο    π    [    ]    ret   del  end  pgdn  kp7  kp8  kp9  kp+
  esc  α    σ    δ    φ    γ    θ    ϑ    κ    λ    ;    '    \                          kp4  kp5  kp6
 @Grek home ζ    χ    ξ    ϕ    β    ν    μ    ,    .    /   @Grek            up         kp1  kp2  kp3  kprt
  lctl lmet lalt           spc                 ralt rmet cmp  rctl       left down rght  kp0  kp.
)
```

slotThe (changes requested):
> IMO something general like this is much preferrable to the current detour via Compose key seqs. After all, that's basically using another configurable keyboard remapper to make up for the shortcomings of this configurable keyboard mapper...

Are you sure that this is universal for X11?  This more seems like something that GNOME or KDE might implement at the DE level.  For example, I tried this just now with both Emacs and alacritty (on xmonad) and neither had support for this feature.  In contrast, compose-key sequences really work everywhere on X11.

leftaroundabout:
Hm, yeah apparently it's rather something that both GTK+ and QT applications do. It does work for basically all applications I ever use (but then, most them actually run inside either xfce4-terminal or Firefox).

Agree that this can't be rolled out as a default behaviour, but it should be a possibility to opt in to, because for the users who do use Gnome / KDE / Mate etc. (which are after all the defaults on most Linux distros), it does save a lot of hassle.

slotThe:
> As this only works in GTK/Qt apps, a better solution is to improve XCompose support. I think this could be done by automatically generating a compose file for all the unicode buttons/chars in `.kbd` file.. With that, hopefully, only thing users would need to do is to `include` the generated compose file in `~/.XCompose`.

This sounds like an interesting idea!  Possibly related: https://github.com/kmonad/kmonad/issues/79

@leftaroundabout I'd be okay with making this an opt-in feature for now so that people who wish can use it.  However, long term, I think going via `XCompose` sounds like the better solution

david-janssen:
Hey everyone,

Sorry I'm a bit late contributing to this conversation. Had a bad flareup and was very under the weather for a week. Still a bit brain-foggy, so if I don't make sense, I blame the allergies ;-).

I'm going to respond to any point that I pickup, even if it's already been addressed by others, just for the sake of clarity.

First of: thanks! I think it's really cool that you want to contribute to KMonad, and I am very happy as well as I can to help answer questions or think things through.

Now, about the implementation details: I think others already mentioned that indeed the ctrl+shift+U method is Window-Managery. I think there was also a method that actually works on the.. uh.. what's it called.. the non-X prompt: virtual console?

> After all, that's basically using another configurable keyboard remapper to make up for the shortcomings of this configurable keyboard mapper...

Although I see your point, I would like to adjust a small detail that makes a world of difference: we're using another configurable keyboard remapper to make up for the shortcomings *of the output signal of keyboard events*. The moment UTF-8 key-event injection becomes a thing in Linux, I'm hopping right on board. The core mission of KMonad is to remap keyboard events to other keyboard events. Everything else is just embroidery. And keyboard events (in linux) use a very small Keycode set. I think Windows actually has native UTF-8 injection? But then you'd be using Windows...

> We can make this opt-in. This indeed make things simple.

Opt-in is not simple for the developer. You'd still be maintaining features (that you might not be using yourself) that only service a subset of your userbase.

> a better solution is to improve XCompose support. 

Perhaps, but on the other hand Wayland is looming. Of course, Wayland has been looming for decades, but it does look like adoption is starting, and these processes go really slowly until they don't, and then they go really quickly.


Actually, I've done a redesign of the configuration-side of KMonad (on the `feature/cfg-refactor` branch). Most of the config-stuff is done, but I still need to do the keymap stuff and hook up the model, so it's not ready, even for a beta build, but it has some features that probably solve this problem in yet another way, and in a way that wouldn't require any extra complexity or programming from our side. 

Let me explain:

One of the big issues we've been dealing with is keycode translation issues between OSes (and also: locale issues). Basically, the correspondence between abstract keycodes (like the 'A-key') and actual numerical representation is very platform dependent. I spent a lot of time working on translation code, but then you really get caught up in looking after all the edge-cases and weird platform-locale interaction differences. So instead, I've decided to make the name-to-code correspondence fully configurable. The new kmonad will look up all the name-to-code correspondences in a Dhall-language configuration file. (Dhall also provides support for imports and functions, so it becomes really easy to keep a clean, modulear config). This should solve all platform/locale issues, since whatever OS or locale you're on, you as the user are now responsible for making sure your names and codes line up, and I will never have to deal with that problem again (hopefully ;)).

Additionally, something that *was* hardcoded but shouldn't be is the name-to-gesture mapping. This is a bit more abstract, but for example, we were parsing 'A' as 'shift-around-a'. That is probably true everywhere, but '%' as 'shift-around-5' is wrong for many locales. So this Dhall also contains a setting that specifies all those correspondences.

This basically already solves the issue, right? You can specify 'Φ' as 'C-S-u 1 2 3 4' (where 1234  should be the actual code for phi) in your dhall config, and now whereever 'phi' occurs in your config, KMonad will read that as that gesture. Then, if you switch window manager to something that doesn't support these sequences, instead you can specify these XCompose sequences. And since Dhall allows imports, you keep 1 file with the compose keys, and another with the CSu keys, and import whichever you need for the platform you're running on. 

Finally, we'll definitely be accepting user contributions to the configuration directory, so you can pull-request in your dhall files with the CSu mappings for greek, and they'll just become part of kmonad, ready to be imported into a cfg file by anyone.

Sorry if this was a bit rambly. Looking forward to feedback.

PR #586, opened 2022-07-15
Support x86_64-darwin in flake

mtoohey31:
This pull request adds support for the `x86_64-darwin` system to the flake, as well as a `list-keyboards` package for that platform. It does the same thing as https://github.com/kmonad/kmonad/pull/566 except for master.

### DIFF
--- a/c_src/mac/.gitignore
+++ b/c_src/mac/.gitignore
@@ -1,0 +1,1 @@
+list-keyboards
--- a/c_src/mac/Makefile
+++ b/c_src/mac/Makefile
@@ -1,2 +1,10 @@
-all: list-keyboards.c
-	gcc $< -o list-keyboards -framework IOKit -framework CoreFoundation
+.PHONY: all install
+
+all: list-keyboards
+
+install: list-keyboards
+	mkdir -p $(DESTDIR)/bin
+	cp $< $(DESTDIR)/bin
+
+list-keyboards: list-keyboards.c
+	$(CC) $< -o $@ -framework IOKit -framework CoreFoundation
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -58,6 +58,18 @@ because you installed it yourself or because you are using NixOS, you can build
 nix-build nix
 ```
 
+On MacOS, you'll have to use something like the following to get nix to pull in
+the karabiner submodule:
+
+```shell
+nix build "./nix?submodules=1"
+```
+
+If you want to pull in kmonad as a flake input for configuring a darwin system,
+you may find it necessary to use a reference like:
+`git+https://github.com/kmonad/kmonad?submodules=1&dir=nix` instead of
+`github:...`.
+
 Another option with `nix` is to use the `nix-shell` to ensure you have the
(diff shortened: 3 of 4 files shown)


MuhammedZakir:
About submodules: executing `git submodule update` as a pre-build hook is better, right? Submodule is necessary when building, so the current suggest way doesn't feel good. :-( Although if that's the only way right now, then leave it as is.

mtoohey31:
> About submodules: executing `git submodule update` as a pre-build hook is better, right? Submodule is necessary when building, so the current suggest way doesn't feel good. :-( Although if that's the only way right now, then leave it as is.

That would be much nicer, but it doesn't look like Nix copies the .git directory to the store when it clones sources, so I don't think that will work. There's two alternatives I can come up with:

1. We add the karabiner submodule as a flake input too and copy it into place during the build. I don't really like this option because it will require us to keep the submodule version in sync with the flake input, and things could get really messy if someone does pull the flake in with submodules.
2. I could add a phase early on that checks if submodules are present when building for darwin. It won't prevent people from having to use the wierd submodule query parameter, but it will at least tell them what to do if they don't use it.

I lean towards the second option, does that sound ok to you?

mtoohey31:
> Sorry for the late response.
> 
> > I lean towards the second option, does that sound ok to you?
> 
> Go ahead with 2nd option. :-)

No worries, added in f25660a4de1b2fa5d09f3ef73d839ba187389ce6!

slotThe:
@MuhammedZakir @mtoohey31 is this in a mergeable state now?  I won't pretend to know about nix on MacOS, so opinions are very welcome :)

mtoohey31:
Yes, it should be good to go. It should be merged at the same time as #566 though, and the way we left things there was that we'd wait for a little bit to see if anyone else had feedback before merging, as per this comment: https://github.com/kmonad/kmonad/pull/566#issuecomment-1185221962. There might not be anyone else to give feedback though; the KMonad + Nix + MacOS niche is probably quite small :slightly_smiling_face:.

rafaelliu:
I have a M1 mac running Monterey and dext, @mtoohey31 fork installed without a flaw. Been running kmonad for a day now, no issues.

Just adding "aarch64-darwin" to `supportedSystems` was enough, should it be added to the PR?

mtoohey31:
Great to hear @rafaelliu, thanks for testing it! I've added that in 1935809d3f1abe9edfce5030969592e59401b72a.

mtoohey31:
I've been using this branch almost daily since I created the PR, and things seem to be working fine. I've just rebased both this branch and the one for #566. Would it be possible to merge these two PRs soon?

slotThe:
@mtoohey31 Yes!  I'll try to get to this before Friday (if not please nag :)

mtoohey31:
Great, thank you!

slotThe:
Thank you!  Sorry for not seeing/getting to this sooner, I've been pretty swamped with other projects

mtoohey31:
No worries, thanks again!

PR #596, opened 2022-07-23
Read from multiple inputs

IzumiRaine:
Hacky implementation of multiple inputs on Linux.

My goal was to use extra buttons on my mouse to switch layers on my keyboard, so I needed a way to attach the same kmonad instance to two input devices. Unfortunately, it turns out the mouse pointer stops moving if I run kmonad on the mouse; so I abandonded the idea (I thought the keyboard and mouse parts would be different endpoints, but apparently not). Maybe this code can still be useful for somebody else as a starting point.

### DIFF
--- a/src/KMonad/Args/Joiner.hs
+++ b/src/KMonad/Args/Joiner.hs
@@ -251,7 +251,7 @@ getCmpSeqDelay = do
 
 -- | The Linux correspondence between IToken and actual code
 pickInput :: IToken -> J (LogFunc -> IO (Acquire KeySource))
-pickInput (KDeviceSource f)   = pure $ runLF (deviceSource64 f)
+pickInput (KDeviceSource f)   = pure $ runLF (deviceSourceMultiple64 f)
 pickInput KLowLevelHookSource = throwError $ InvalidOS "LowLevelHookSource"
 pickInput (KIOKitSource _)    = throwError $ InvalidOS "IOKitSource"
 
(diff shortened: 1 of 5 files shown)


jtroo:
> the mouse pointer stops moving if I run kmonad on the mouse

Without having looked into the code specifics, I think this is fixable this by forwarding all received events that kmonad doesn't care to process to uinput.

There may be some extra code that needs to be added to uinput registration for the kernel to accept kmonad's forwarded mouse events, though I'm not sure about that.

MuhammedZakir:
> > the mouse pointer stops moving if I run kmonad on the mouse
> 
> Without having looked into the code specifics, I think this is fixable by forwarding all received events that kmonad doesn't care to process to uinput.

Yes.

> There may be some extra code that needs to be added to uinput registration for the kernel to accept kmonad's forwarded mouse events, though I'm not sure about that.

If I understand correctly, in this line [1], changing `256` to `275`[2] should allow sending events for commonly used mouse buttons.

[1] https://github.com/kmonad/kmonad/blob/95bb58b7347ebacb4654fb21c798e1cd10331b0d/c_src/keyio.c#L21
[2]  These are keycodes; see `/usr/include/linux/input-event-codes.h`.

slotThe:
@IzumiRaine any updates on this?

IzumiRaine:
Sorry, nothing yet. I'll take another look around October (will be a bit too busy before then). Feel free to close the PR meanwhile.

Ninlives:
@IzumiRaine Hi, could you please present a small configuration example of how to use multiple inputs? Thanks!

IzumiRaine:
@Ninlives sure. You can add multiple paths to `device-file` like this:
```
(defcfg
  input  (device-file "/dev/input/by-id/keyboard1" "/dev/input/by-id/keyboard2")
  output (uinput-sink "My KMonad output" "")
)
```
But please keep in mind that so far this is just a quick hack and only for Linux.

stevep99:
+1 on this feature.

It would be nice as well if you could specify "*" as the input device, in then it would match _all_ attached keyboard devices.

PR #608, opened 2022-08-19
Add OpenRC service

alternateved:
Hello!

Since there already is support for both systemd and runit, I would also like to add openrc to that bunch. I tried to mirror here the already great instructions available in runit service.

Best wishes,
alternateved


slotThe:
Thanks!

PR #615, opened 2022-09-04
add aliases for "scroll up/down"

tholzschuh:
Added aliases for "Scroll Down" and "Scroll Up".

Feel free to replace them by any better alias, they were just the first I came up with.
But as my keyboard has these two keys, I'd really like to have an alias for them.


slotThe:
Looks fine to me; thanks!

PR #632, opened 2022-10-04
Include security option information for `podman` users

bryce-carson:
Without including this option, the `/host/` directory is inaccessible because of SELinux.


slotThe:
Thanks!

PR #639, opened 2022-10-26
FAQ: how to easily select Linux keyboard device with a shell script

pcppcp:
I keep moving my laptop around the apartment and I constantly unplug and re-plug external keyboard to two different computers.
In order to keep the same layout everywhere, I've made a simple script that makes selection of the input device easier and I thought I'd share it in case someone might find it useful.

PS: @david-janssen thanks for this wonderful and truly lifechanging app!


slotThe:
This is neat, thank you!

PR #670, opened 2023-01-02
feat: add tap-next-press

JakobBruenker:
This addresses the question I posed in https://github.com/kmonad/kmonad/discussions/656.

tap-next-press acts just like tap-next-release, except that it acts as hold on key presses rather than on key releases.

I suppose there could be one or two more functions added to fill out the matrix with `hold` and such, but I'd rather find out whether this has a decent chance of getting merged before I implement those, since I don't personally have a use for them.

### DIFF
--- a/doc/quick-reference.md
+++ b/doc/quick-reference.md
@@ -271,6 +271,13 @@ to be the most comfortable.
   (defalias tnr (tap-next-release a sft))
   ```
 
++ `tap-next-press`: like `tap-next` but decide whether to tap or hold
+  based on whether another key is pressed before this one is released.
+
+  ```clojure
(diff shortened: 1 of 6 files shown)


slotThe (commented):
This looks fine to me, thanks!  I think we can merge this after some docs are added to the tutorial/the quick reference

> I suppose there could be one or two more functions added to fill out the matrix with `hold` and such, but I'd rather find out whether this has a decent chance of getting merged before I implement those, since I don't personally have a use for them.

I'd wait with these until someone actually has a real use-case for them/requests the feature

JakobBruenker:
Thanks @slotThe, I've added documentation

slotThe:
Great, thanks!

PR #690, opened 2023-03-28
Fix doc comment by substituting "sink" for "source" in IO.hs

gwimm:
Falsely labeled sources as sinks in doc comments. This pull request simply substitutes `sink` for `source` in the doc comments for `mkKeySource`.


slotThe:
Thanks!

PR #694, opened 2023-04-23
Making systemd.service a template to support multiple keyboards

nharward:
# Multiple keyboard support for systemd

Updates the systemd service file to be a template file. Rather than only
support `/etc/kmonad/config.kbd`, the template supports
`kmonad@<kbd-name>.service`, where `name` corresponds to `/etc/kmonad/kbd-name.kbd`.
If `<kbd-name>` is not given a value of `config` is used.


slotThe:
Thanks!

PR #697, opened 2023-05-06
systemd.unit support for filenames with hyphens

nharward:
When using `/etc/kmonad/*.kbd` files with hyphens in the name systemd
will escape them to slashes when using `%I`. This commit changes to use
`%i` instead which does not escape the name, resulting in the correct
filename passed to kmonad.

It's important to use `%i` for the filename so that multiple instances
of kmonad can run for different keyboards at the same time, say for a
laptop and external keyboard. The `DefaultInstance=config` in line 11
will use a default value of `config` if no other value is given.

Examples:

    `systemctl start kmonad@foo`     # Runs `kmonad /etc/kmonad/foo.kbd`
    `systemctl start kmonad@foo-bar` # Runs `kmonad /etc/kmonad/foo-bar.kbd`
    `systemctl start kmonad@`        # Runs `kmonad /etc/kmonad/config.kbd`

### DIFF
--- a/startup/kmonad@.service
+++ b/startup/kmonad@.service
@@ -4,8 +4,9 @@ Description=kmonad keyboard config
 [Service]
 Restart=always
 RestartSec=3
-ExecStart=/usr/bin/kmonad %E/kmonad/config.kbd
+ExecStart=/usr/bin/kmonad %E/kmonad/%i.kbd
 Nice=-20
 
 [Install]
+DefaultInstance=config
 WantedBy=default.target


nharward:
Looks like the failed workflow step was on a failure of GHC to Windows. Is it possible to re-trigger this flow to see if it was intermittent, or push through regardless? This PR is only for a contrib file related to systemd and does not affect kmonad core code.

slotThe:
Oh, sorry, I must have forgotten about this. AFAIR the workflow is due to some upstream issue; should go away by itself

Thanks!

PR #747, opened 2023-10-16
Update nixos-module.nix

AndreyAdrian:
flipping serviceConfig.Restart to "yes"

In nixos theres problems because the service is not restarting by default, and have to unplug/replug, or completely reboot the computer.

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -126,7 +126,7 @@ let
       value = {
         description = "KMonad for ${keyboard.device}";
         script = lib.escapeShellArgs cmd;
-        serviceConfig.Restart = "no";
+        serviceConfig.Restart = "yes";
         serviceConfig.User = "kmonad";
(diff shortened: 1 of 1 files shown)


slotThe:
I suppose this should be free of unintended consequences—thanks!

jooooscha:
I get the following error on rebuilding my config:

```
error: A definition for option `systemd.services.<...>.serviceConfig' is not of type `attribute set of (systemd option)'. Definition values:
       - In `/nix/store/dg9njr183vi6hmayn56wk7ls203yr8ki-source/nix/nixos-module.nix':
           {
             Nice = -20;
             Restart = "yes";
             SupplementaryGroups = [
               "input"
           ...
```

I think "yes" is not a valid value for the `Restart=` config option. See [man](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=)

Instead of `yes` this should be changed to `always`

PR #762, opened 2023-11-02
keycodes: Add aliases for `copy` and `pste` for KeyCopy and KeyPaste

markstos:
These are useful for creating universal copy/paste shortcuts that work
inside and outside of terminals and potentially across operating
systems.

https://github.com/kmonad/kmonad/discussions/461#discussioncomment-7452065

### DIFF
--- a/src/KMonad/Keyboard/Keycode.hs
+++ b/src/KMonad/Keyboard/Keycode.hs
@@ -382,6 +382,8 @@ aliases = Q.mkMultiMap
   , (KeyScrollLock,     ["scrlck", "slck"])
   , (KeyScrollUp,       ["scrup", "sup"])
   , (KeyScrollDown,     ["scrdn", "sdwn", "sdn"])
+  , (KeyCopy,           ["copy"])
+  , (KeyPaste,          ["pste"])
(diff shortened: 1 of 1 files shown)


src/KMonad/Keyboard/Keycode.hs
```diff
@@ -382,6 +382,8 @@ aliases = Q.mkMultiMap
   , (KeyScrollLock,     ["scrlck", "slck"])
   , (KeyScrollUp,       ["scrup", "sup"])
   , (KeyScrollDown,     ["scrdn", "sdwn", "sdn"])
+  , (KeyCopy,           ["copy"])
+  , (KeyPaste,          ["pste"])
```
slotThe, line 386: For a key named `KeyName` we automatically generate an alias of `name`, so `copy` is superfluous here. `pste`, however, seems fine.

slotThe (changes requested):
F

src/KMonad/Keyboard/Keycode.hs
markstos: Thanks, I think I'll just use `paste` as the alias.

PR #763, opened 2023-11-02
keycodes: alias KeyMenu to menu

markstos:
This keycode opens what's often the "right click context menu" in apps
that support it.

It's useful in workflows like this:

You type something and then an underline appears to indicate it was
misspelled. You can use the arrow keys to go back to the word, then use
the "menu" key to open the menu, select the correct word and press Enter
to accept.

This key is rarely used because it's not conventionally located on
keyboards, but with KMonad, it can be mapped somewhere convenient on a
layer!

### DIFF
--- a/src/KMonad/Keyboard/Keycode.hs
+++ b/src/KMonad/Keyboard/Keycode.hs
@@ -395,6 +395,7 @@ aliases = Q.mkMultiMap
   , (KeyComma,          ["comm", ","])
   , (KeyDot,            ["."])
   , (KeySlash,          ["/"])
+  , (KeyMenu,           ["menu"])
   , (KeyNumLock,        ["nlck"])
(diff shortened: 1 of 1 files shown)


slotThe:
Related: https://github.com/kmonad/kmonad/pull/762

> For a key named `KeyName` we automatically generate an alias of `name`

So `menu` should already be taken care of.

markstos:
Got it. Thanks for the reply. 

Something like `kmonad --list-aliases` would be helpful to make it easier to discover what aliases are supported.

slotThe:
> Something like `kmonad --list-aliases` would be helpful to make it easier to discover what aliases are supported.

That's true; it's not super obvious that this autogeneration happens (as evidenced by this PR, I suppose)

Fancy another contribution? :)

PR #766, opened 2023-11-06
Update keycap name for non-US backslash in Mac/Types.hs

postylem:
Change `KeyNonUSBackslash` to `Key102nd` to match the name defined in `src/KMonad/Keyboard/Keycode.hs` (they keycode which is assigned alias 'nubs').

Since `Key102nd` is assigned the the 'nubs' keycode (aka '102d', or 'lsgt') in Keycode.hs, it seems it should also be the keycode to which mac keycode (0x7, 0x64) (that is, the <kbd>§</kbd> key to the left of <kbd>1</kbd> on a [UK-layout ISO keyboard](https://support.apple.com/en-us/HT201794)) is assigned.


src/KMonad/Keyboard/IO/Mac/Types.hs
```diff
@@ -214,7 +214,7 @@ kcMap = M.fromList
   , ((0x7,0x61), KeyKp9)
   , ((0x7,0x62), KeyKp0)
   , ((0x7,0x63), KeyKpDot)
-  -- , ((0x7,0x64), KeyNonUSBackslash)
+  -- , ((0x7,0x64), Key102nd)
```
slotThe, line 217: This is commented out; for a good reason I presume. Is this even a valid keycode on Mac?

postylem: I assume it's commented out because it's designed for the ANSI keyboard that comes on macs sold in the US. But macs sold in Europe (and some other places) have an ISO layout keyboard, which has one additional key. 

The additional key is mac key code (7,100). I have a macbook pro with an ISO keyboard. When I pressed it, monad would crash. 

In order to use this key, I needed to uncomment this line, but also had to change it from KeyNonUSBackslash to Key102d. This pull request would remove that second step for anyone else with such a keyboard.

slotThe: Ah, I see; thanks for explaining!

> I don't know whether perhaps it would cause some problem for folks with an ANSI keyboard to have this uncommented (if not then uncommenting would be great!)

I… really hope not :) This would mean we're doing something evil somewhere, which would be a bug. Feel free to uncomment the line and then we can merge this

postylem: Right, I can't see how it would cause a problem to have extra un-triggerable keycommands in this list. I certainly hope that's the case.

I uncommented it.

slotThe:
Thanks!

postylem:
I also had to modify the keymap/template/apple.kbd to work with my ISO macbook pro (shifting things over and adding this one additional key). Would it be helpful to submit a PR an keymap/template/apple-ISO.kbd ?

tjwelde:
I think that would be very helpful.

Issue #753, opened 2023-10-17
PC wakes up keyboard not working


CyprienRicque:
Hello,
When the PC wakes up from sleep, sometimes the keys are correctly received by kmonad (displayed with -ldebug) but they are not transmitted back to the PC. So I can't use the keyboard anymore.

## Config
macOS Sonoma 14.0
M1 Pro

nightscape:
Same here. I noticed that you can make kmonad crash by pressing the `^` key left of `1`.
I've built a launcher script around kmonad that restarts it after a crash:
```bash
#!/bin/bash

# Path to the kmonad binary
KMONAD_PATH="/Users/martin/.local/bin/kmonad"
KMONAD_CONFIG_PATH="/Users/martin/Workspaces/devtools/kmonad-contrib/keymaps/nightscape/bone_mac.kbd"

# Counter for how many times kmonad has crashed
CRASH_COUNT=0

# Timestamp of the first crash in the current sequence
FIRST_CRASH_TIMESTAMP=0

while true; do
    # Run kmonad with superuser permissions
	echo "Starting kmonad"
	echo "sudo $KMONAD_PATH $KMONAD_CONFIG_PATH"
    sudo $KMONAD_PATH $KMONAD_CONFIG_PATH

    # If kmonad exits, it has crashed. Increment the crash counter.
    CRASH_COUNT=$((CRASH_COUNT + 1))

    # If this is the first crash in the current sequence, record the timestamp
    if [ $CRASH_COUNT -eq 1 ]; then
        FIRST_CRASH_TIMESTAMP=$(date +%s)
    fi

    # Get the current timestamp
    CURRENT_TIMESTAMP=$(date +%s)

    # Calculate the time difference between the first crash and now
    TIME_DIFF=$((CURRENT_TIMESTAMP - FIRST_CRASH_TIMESTAMP))

    # If kmonad has crashed 3 times in less than 60 seconds, exit the loop
    if [ $CRASH_COUNT -ge 3 ] && [ $TIME_DIFF -lt 60 ]; then
        echo "kmonad has crashed too frequently. Exiting."
        exit 1
    fi

    # If more than 60 seconds have passed since the first crash, reset the crash counter
    if [ $TIME_DIFF -ge 60 ]; then
        CRASH_COUNT=0
    fi

    # Wait for a second before trying to restart kmonad
    sleep 1
done
```

[...]

PR #781, opened 2023-12-27
Mac: ignore keyboards named "Karabiner ..."

brttbndr:
This PR addresses https://github.com/kmonad/kmonad/issues/753

# Problem

In config, you can optionally specify a keyboard name:

```
input (iokit-name "Apple Internal Keyboard / Trackpad")
input (iokit-name) ;; no name specified
```

If omitted, then KMonad will attach to any keyboards it can find, skipping those that are provided by the Karabiner VirtualHID driver. However -- this filtering is performed by exact string comparison, and the name of the device varies by driver version.

When the VirtualHID device is not filtered out, a loop results; events emitted by KMonad are received by KMonad, processed again, and so on ...

This might look like the following (for a single initial tap on `a`). Note that the `Emitting: Press <a>` is followed by `Received event: Press <a>` (i.e. looped).

```
Received event: Press <a>
Running hooks
Block level set to: 1
Registering untimed hook: 1993
Registering 300ms hook: 1992

Received event: Release <a>
Running hooks
Emitting: Press <a>
Unblocking input stream, no stored events
Emitting: Release <a>

Received event: Press <a>
Running hooks
Block level set to: 1
Registering untimed hook: 1995
Registering 300ms hook: 1994
```

# Fix

Filter out keyboards prefixed "Karabiner " which matches across both older and current (DriverKit-based) versions of the VirtualHIDKeyboard device

## Examples of matched device names

Only the first is filtered by existing code; any / all of the 3 are filtered with this PR.

```
Karabiner VirtualHIDKeyboard
Karabiner DriverKit VirtualHIDKeyboard 1.6.0
Karabiner DriverKit VirtualHIDKeyboard 1.7.0
```

### DIFF
--- a/c_src/mac/keyio_mac.hpp
+++ b/c_src/mac/keyio_mac.hpp
@@ -75,7 +75,7 @@ void open_matching_devices(char *product, io_iterator_t iter) {
             return;
         }
     }
-    CFStringRef cfkarabiner = CFStringCreateWithCString(kCFAllocatorDefault, "Karabiner VirtualHIDKeyboard", CFStringGetSystemEncoding());
+    CFStringRef cfkarabiner = CFStringCreateWithCString(kCFAllocatorDefault, "Karabiner ", CFStringGetSystemEncoding());
     if(cfkarabiner == NULL) {
         print_iokit_error("CFStringCreateWithCString");
         if(product) {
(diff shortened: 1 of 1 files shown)


slotThe:
Thanks!

PR #807, opened 2024-01-29
Work around fn key issue

pfgithub:
When pressing the fn key, the application crashes

```
kmonad: toEnum{Keycode}: tag (464) is outside of enumeration's range (0,255)
```

#372 should fix this for real but until then this makes sure the application doesn't crash when pressing the fn key


slotThe:
Seems like a reasonable temporary fix; thanks!

PR #813, opened 2024-02-10
Fix broken kmonad-bin nixos link

znd4:
I'm not sure if I got the SHA right (`nix-prefetch-url`, right?)


slotThe:
> I'm not sure if I got the SHA right (`nix-prefetch-url`, right?)

I believe so, yes.

Thanks!

Issue #811, opened 2024-02-04
[FR] Arabic character support


mdSlash:
KMonad lacks support for Arabic Unicode. Would you please consider adding Arabic support in the near future?

mdSlash:
Hey @slotThe, thanks for this awesome project. I wanna contribute by adding Arabic characters. What do I need to do to achieve that? Do I just need to add the characters to the "ComposeSeq.hs" file, or is there something else I should do?

slotThe:
@mdSlash I don't really know how Arabic characters work, but if they can be composed using X11 compose key sequences then that file is exactly the right place to add them!

mdSlash:
It doesn't necessarily require inserting Arabic characters. I'm using `setxkbmap` to change the layout to Arabic and adding English characters with their Arabic equivalents in the KMonad config.

PR #814, opened 2024-02-11
Add Arabic chars

mdSlash:


mdSlash:
#811

slotThe:
@mdSlash as you can see from the CI, it seems like GHC does not like some Unicode endpoints and you'll have to use the appropriate escape sequence instead

mdSlash:
@slotThe I tried to fix the Unicode problem. I hope it's working now!

Will it be possible to use Arabic chars inside layers without showing a `Parse error` message, without needing to use compose-key sequences in the `tap-macro` functions?

slotThe:
> Will it be possible to use Arabic chars inside layers without showing a `Parse error` message, without needing to use compose-key sequences in the `tap-macro` functions?

You tell me, it's your PR! :) I would think so, but this should obviously be tested before this is merged

mdSlash:
@slotThe I'm stuck here and have no idea what to do next. You can close this pull request if you're not making any edits to it. Thanks.

mdSlash:
> Try `\x` instead of `\u`

I will try it. Thanks for your help.

PR #826, opened 2024-03-20
Simplify NixOS installation instructions

Cre3per:
I configured KMonad on NixOS yesterday. I struggled because there were multiple installation/configuration methods. In the end I found that there is an up-to-date KMonad package on NixOS and half of my configuration was already taken care of by the provided NixOS module.

The new instructions

- are updated (There is a KMonad package on NixOS)
- use examples rather than wildcards to be more friendly to NixOS beginners
- remove alternative installation/configuration instructions to be more streamlined

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -384,117 +384,46 @@ You can install `kmonad` via `xbps-install`:
 
 ### NixOS
 
-There is not currently a `kmonad` package in `nixpkgs`, however the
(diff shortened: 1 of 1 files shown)


slotThe (commented):
Thanks! I feel like at this point it would be most useful to instruct people on how to include kmonad as a flake input so they don't have to manually clone the repository somewhere

slotThe:
Thanks!

PR #831, opened 2024-04-03
Added named source blocks

jokesper:
This PR adds named `defsrc` blocks.
To keep backwards compatibility I decided to name them `defsrc'`.

There are some design decisions which could be changed:
1. `defsrc'` instead of something else like `defsrc-named`.
  But I found finding the right name is hard. :|
2. `defsrc'` instead of expanding `defsrc`.
  Expanding `defsrc` through sub expressions
  would be possible since no key could be a sub expression.
  Furthermore it would allow future expansion (e.g.: fallthrough per source)
  ```
  (defsrc (name qwerty)
    q w e r t y
  )
  ```
I opted for `defsrc'` as it was the easiest. Also note that keys not defined in this source are passthrough to other layers (`_`). This was simply the easiest way to implement.
Let me know if you find some issues.

### DIFF
--- a/src/KMonad/Args/Parser.hs
+++ b/src/KMonad/Args/Parser.hs
@@ -368,10 +368,17 @@ defaliasP = many $ (,) <$> lexeme word <*> buttonP
 -- $defsrc
 
 defsrcP :: Parser DefSrc
-defsrcP = many $ lexeme keycodeP
-
+defsrcP =
+  DefSrc
+    <$> optional (keywordP "name" word)
+    <*> many (lexeme keycodeP)
 
 --------------------------------------------------------------------------------
 -- $deflayer
+
 deflayerP :: Parser DefLayer
-deflayerP = DefLayer <$> lexeme word <*> many (lexeme buttonP)
+deflayerP =
+  DefLayer
+    <$> lexeme word
+    <*> optional (keywordP "source" word)
+    <*> many (lexeme buttonP)
--- a/doc/quick-reference.md
+++ b/doc/quick-reference.md
@@ -68,7 +68,7 @@ The following are all global config options that one can set in the
 `defcfg` block.
 
 + `fallthrough` (boolean, defaults to `false`): re-emit keys that are
-  not defined in a `defsrc` block.
+  not defined in the corresponding `defsrc` block.
 
   This allows one to only specify certain parts of a layout, with all
   other keys having their "default" meaning.
@@ -317,6 +317,8 @@ just collections of keys.
   `defsrc` is very similar to a layer visually, it is not a one and will
   thus not be used as one! It only serves to define where the different
   keys are and what kind of layout kmonad is initially dealing with.
+  It also supports giving a name via `:name <my-soucre-name>` as the
+  first argument.
 
   For example, an ANSI 60% keyboard may be represented as:
 
@@ -337,6 +339,9 @@ just collections of keys.
 
    For example, defining a qwerty layer, as well as one for special
    symbols and numbers:
+
+   To use a named source block add `:source <my-source-name>` after
+   the layer name.
 
   ```
   (deflayer qwerty
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -61,9 +61,9 @@
 
     KMonad catches input events and tries to match them to various handlers. If
     it cannot match an event to any handler (for example, if it isn't included
-    in the `defsrc` block, or if it is, but the current keymap does not map any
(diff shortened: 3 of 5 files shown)


slotThe (commented):
This looks great, thanks!

> Expanding defsrc through sub expressions would be possible since no key could be a sub expression. Furthermore it would allow future expansion (e.g.: fallthrough per source)

I think I would favour this solution.

Syntax-wise, perhaps something this would fit:

``` clojure
(defsrc {:name my-name}
  …)
```

(this is akin to how Clojure functions can be given an attribute set with some metadata information)

jokesper:
Wouldn't something like
```
(defsrc (name my-name)
 ...)
```
make more sense.
I.e. let defsrc take not just keycodes but also defsrc options. This wouldn't introduce new syntax.
The only difference would be that it might be unclear what you mean, when applied to other functions

slotThe:
I figured that `{:name my-name}` might better communicate the intent of these being additional options, though I don't think there's anything wrong with `(name my-name)` either. No strong opinions on my side; pick whichever form you like best.

jokesper:
Wouldn't something like
```
(defsrc (name my-name)
 ...)
```
make more sense.
I.e. let defsrc take not just keycodes but also defsrc options. This wouldn't introduce new syntax.
The only difference would be that it might be unclear what you mean, when applied to other functions

slotThe:
> Wouldn't the braces cause problems, since they aren't keywords and would reference buttons?

This parser would run before the button parser, so this should not cause any issues. Instead of the current

``` haskell
  , try (symbol "defsrc") *> (KDefSrc <$> defsrcP)
```

we would have something like

``` haskell
  , try (symbol "defsrc") *> (KDefSrc <$> defsrcAttrSetP <*> defsrcP)
```

which would not cause any ambiguities

jokesper:
True, but It would be best for the attribute set to be optional to allow for backwards compatibility.
But something like:
```
(defsrc { } )
```
Which would previously define the source keys to be the braces, would now be an empty source with an attribute set.
with parenthesis this doesn't happen, since they already have to be escaped.

jokesper:
I was just looking at the code and noticed how `tap-macro` handles it.
Maybe something like
```
(defsrc :name my-name
    a b c)
```
would be ideal.
So simply leave out the braces and signal the optional argument via the colon from `keywordP`

slotThe:
> True, but It would be best for the attribute set to be optional to allow for backwards compatibility. But something like:
>
> ```
> (defsrc { } )
> ```
>
> Which would previously define the source keys to be the braces, would now be an empty source with an attribute set. with parenthesis this doesn't happen, since they already have to be escaped.

Seems more logical to me to simply not parse empty attribute blocks. Either way, I also think that

``` clojure
(defsrc :name my-name
    a b c)
```

looks fine.

jokesper:
I implemented the changes. Please take another look.

src/KMonad/Args/Parser.hs
```diff
@@ -367,11 +367,25 @@ defaliasP = many $ (,) <$> lexeme word <*> buttonP
...
-defsrcP = many $ lexeme keycodeP
+newSettings :: Monoid s => ASetter' s [a] -> a -> s
+newSettings field value = set field [value] mempty
 
+defsrcP :: Parser DefSrc
+defsrcP = let f s v p = keywordP s $ newSettings v <$> p in
+  fmap fold . many . lexeme . choice . map try $
+    [ f "name" name word
+    , newSettings keycodes <$> keycodeP
+    ]
```
slotThe, line 378: This seems a bit "unnecessarily general", I think. To me it seems more sensible to restrict `defsrc` blocks to only have one name, specified right after the `defsrc` keyword (this would incidentally also make the implementation more straightforward)

jokesper: I used it since when I implemented it I also added `fallthrough` as an option. Removing it would simplify the code though it would be readded when imlpmenting other options. Furthermore it adds the ability to pick where you want to add it (start, end, or even in the middle).
Could still remove it if code simplicity is the goal.

slotThe: > Removing it would simplify the code though it would be readded when imlpmenting other options.

We can certainly re-add it when other options come into play, but this does not seem especially relevant for now.

> Furthermore it adds the ability to pick where you want to add it (start, end, or even in the middle). This is nice since in ather cases it's optionally at the end.

I think that this is actually a disadvantage—don't want to give people too much room to mess things up :) Having it always be at the start of the expressions seems fine to me.

jokesper:
I applied your suggested changes. Please take a look again.

slotThe:
Thank you!

Issue #822, opened 2024-03-09
Using tap-hold, no output with multiple modifiers and 1 key tapped


IanEdington:
I have 3 tap-hold keys in my setup, space and the two keys on either side. All the modifiers work with all the other keys, and a single tap-hold modifier works with any of the other tap-hold key. However, holding two tap-hold modifiers and tapping the 3rd tap-hold key doesn't do anything.

using `evtest` with the "My Kmonad Output" I get:

| Hold | Tap | Expected | Actual |  | 
|--------|--------|--------|--------|--------|
| @met_ret | @met_spc | met+spc | met+spc | :white_check_mark: |
| @met_spc | @met_ret | met+ret | met+ret | :white_check_mark: ||
| @met_ret+@ctl_esc | a | met+ctl+a | met+ctl+a | :white_check_mark: |
| @met_ret+@ctl_esc | @met_spc | met+ctl+spc | met+ctl | :x: |

```haskell
(defcfg
  output (uinput-sink "My KMonad output")
  cmp-seq lalt
  cmp-seq-delay 5
  fallthrough true
  allow-cmd true
)

(defsrc lalt spc ralt)

(defalias
  ctl_esc (tap-hold-next-release 200 esc lctl)
  met_spc (tap-hold-next-release 200 spc rmet)
  met_ret (tap-hold-next-release 200 ret lmet)
)

(deflayer qwerty @met_ret @ctl_esc @met_spc)
```

bonsaiiV:
Are you sure, that the issue is not with the maximum key-rollover of your keyboard?
It is common, that certain keys can not be pressed together.
You can test that online for example:
https://www.mechanical-keyboard.org/key-rollover-test/
However this doesn't work if your the keys are captured by your browser.
You can also try whether the keys work together while kmonad is not running to check whether the issue is with your keyboard or kmonad.

IanEdington:
Without kmonad it works, with kmonad it doesn't. That doesn't sound like a key-rollover issue to me, but I know very little about the internals of keyboards.

Happy to provide additional info if you're interested in investigating or if it comes up with someone else.

Issue #764, opened 2023-11-04
Home row mods M-S-], C-M-<Backspace> missing?


brosasaki:
After setting up home row modifiers keys via e.g. `C/a (tap-hold-next-release 200 a lctl)`, KMonad works excellently!
Save for one issue: M-S-] (AKA Gui-Shift-], Command-}, Super-}, etc.), in particular when using the left home-row-mods, does not pass through, despite the normal right modifiers, left modifiers, and right home-row-mods working perfectly with it. I've checked `sudo kmonad config.kbd -l debug` and the log shows KMonad recognizing the meta and shift layers from the home row mods and all other keys except for `]`; debug simply doesn't output anything at all upon press. I've looked my config up and down, but the odd, asymmetrical nature of the issue makes me suspect that it's a KMonad issue, and not a config or Firefox (which uses Command-Shift-] for moving right one tab) issue.
My system is a MacBook Pro 12,1 running macOS Monterey 12.6.8, and I am using the vosaica/kmonad branch and dext version 2.1.0 according to [amiorin's guide](https://gist.github.com/amiorin/4c74f63fe599a1dcbd0933628df1aac9). It appears that the fork has been merged, but I couldn't get kmonad/kmonad to work with a dext 2.1.0 from a Karabiner-Elements install (now removed) and the documentation doesn't reflect the merged branch; LMK if there is a way to use the features from vosaica/kmonad in the main branch, and I can try and make a PR for the documentation.
Here's my config, for good measure:
```
;; See keymap/tutorial.kbd for a more complete tutorial

#|
(deflayer template
  _    _    _    _    _    _    _    _    _    _    _    _    _   
  _    _    _    _    _    _    _    _    _    _    _    _    _    _
  _    _    _    _    _    _    _    _    _    _    _    _    _    _
  _    _    _    _    _    _    _    _    _    _    _    _    _
  _    _    _    _    _    _    _    _    _    _    _    _    _
  _    _    _    _              _              _    _    _    _    _   )

|#

;;;; Setup
(defcfg
  input (iokit-name "Apple Internal Keyboard / Trackpad")
  output (kext) ;; shouldn't this be called "dext" now?
  fallthrough true
  allow-cmd true)

(defsrc
  esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12
  grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc
  tab  q    w    e    r    t    y    u    i    o    p    [    ]    \
[...]

PR #855, opened 2024-06-29
Added key rollover to FAQ

jokesper:
This issue appeared at least twice in #764 and #822.
I think it's a good idea to add it to the FAQ


jokesper:
CI failure for a readme change... nice; espacially the reason
```
file returned by server too large: <repo>/01-index.tar.gz (expected exactly 123043592 bytes)
```

slotThe:
Thanks!

PR #859, opened 2024-07-09
Implemented `around-only` and `around-when-alone`.

### LINKED ISSUE #733: Map to a special symbol (e.g. "_") without holding shift
According to the [tutorial](https://github.com/kmonad/kmonad/blob/master/keymap/tutorial.kbd#L381-L384),

> We have already seen other examples of modded buttons, \\(, \\), *, and +. There are no Keycodes for these buttons in KMonad, but they are buttons. They simply evaluate to `(around lsft x)`. All shifted numbers have their corresponding characters, the same is true for all capitals, and < > : ~ " | { } \\_ + and ?.

However, this causes some inconvenience and slows down my typing compared to QMK. The reason is that
1. If I try to type `a_b` but press `b` before releasing the key that mapped to `_`, it would become `a_B`.
2. More interestingly, if I actually try to type `a_BCD` but release `_` after I press and hold shift, then the shift is released even if I'm holding it, and I eventually get `a_bcd`.

To put it in the Kmonad language:

- We have some button `_`
- `_` is bound to '_' on the input keyboard
- the letters a, lsft are bound to themselves
- Px signifies the press of button x on the keyboard
- Rx signifies the release of said button
- Tx signifies the sequential and near instantaneous press and release of x
- 100 signifies 100ms pass

What it currently does:

    P_ Pa R_ Ra           -> _A 
    P_ Plsft R_ Ta    -> _a

What I expect it to do (so I can type the next key without worrying about releasing the previous key):

    P_ Pa R_ Ra           -> _a
    P_ Plsft R_ Ta    -> _A

~~I have a feeling that I might need to use the compose-key sequence but the [tutorial](https://github.com/kmonad/kmonad/blob/master/keymap/tutorial.kbd#L910-L960) might need some update. The example config doesn't match the document.~~

### LINKED ISSUE #840: "around" mapping that releases the modifier immediately
On an unmodified keyboard, modifier keys can be released separately from the modified key. In other words, I can do this:

```
press shift
press a
release shift
release b
release a
```
This types `Ab`.

However if I use kmonad to map something else to `A`, then they are always released together. Assume I mapped `a` to `A`. Then this

```
press a
press b
release a
```
means
```
press shift
press a
press b
release b
release shift
release a
```
This types `AB`.

I have a layer with a lot of symbols and I type quickly, so I am making a lot of typos with capital letters after symbols due to this issue.

Is there any way to map `a` to `A`, *but* release the virtual shift key once the `A` is emitted, so that I can recover the `Ab` behavior?

jokesper:
This fixes #733 and also fixes #840.
For details on what has been done see the changes in the docs.

### DIFF
--- a/src/KMonad/Model/Button.hs
+++ b/src/KMonad/Model/Button.hs
@@ -39,6 +39,8 @@ module KMonad.Model.Button
 
   -- * Button combinators
   -- $combinators
+  , aroundOnly
+  , aroundWhenAlone
   , aroundNext
   , aroundNextTimeout
   , aroundNextSingle
@@ -65,6 +67,7 @@ import KMonad.Model.Action
 import KMonad.Keyboard
 import KMonad.Util
 
+import qualified RIO.HashSet as S
 
 --------------------------------------------------------------------------------
 -- $but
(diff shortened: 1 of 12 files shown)


jokesper:
I have now rebased the changes and resolved the merge conflict.

src/KMonad/Model/Button.hs
```diff
@@ -191,6 +194,65 @@ around outer inner = Button
...
+ where
+  go :: KeyPred -> AnyK ()
+  go isMyRelease = hookF InputHook $ \e ->
+    if
+      | isMyRelease e -> do
+        runAction $ inner^.releaseAction
+        runAction $ outer^.releaseAction
+        pure Catch
+      | isPress e -> do
+        runAction $ outer^.releaseAction
```
jokesper, line 217: When having to buttons `A` and `B` and they are pressed in sequence, we have the following output
`Psft Pa Rsft Psft Pb`
Previously I thought this wasn't an issue since pressing and releasing modifier keys is mostly irrelevant except for games, but there you wouldn't want complex buttons.
But there are two modifiers for which this is only partially true: `super` and `alt` since a tap `super` may result in a menu opening and `alt` is commonly used to focus the nav bar of an application.

jokesper: I tested Firefox and libre office with regard to pressing alt another key and repressing alt.
In LibreOffice everything works as expected, since the bar is active when alt is held. In firefox it vanishes on the second press as alt toggles the bar.
Fixing this would require a special button which waits for the next event and in case it is its own release annihilates both outgoing events. This would require some changes to the kmonad core event handler.

jokesper: An `implicit-around` of some modifier key, which is not shift and one also wants to hold and press other keys during this time, if used at all would probably be used for a button merging modifiers.
If we really want to not break this behaviour then I think the commit, which changes said default could be left out and we could make a note to change this in the next major release.

slotThe: One may or may not call this expected behaviour, as well—the button *does* say that it re-presses the modifier, after all :)

jokesper: Hmmm. Still the issue (?) with changing the default remains.

slotThe: Mh, I guess you're right that it wouldn't be super nice to change the default in a minor release. On the other hand, I don't think this will affect that many people, so it might be fine

jokesper:
Hmm, there seem to be many issues in my code style which could have been solved by a formatter and maybe a spell / grammar checker (#866 - whoops).
As I have seen different code formatting choices used throughout the code base I was wondering which I should use for KMonad PRs.

slotThe:
Most of it is just trailing whitespace and grammar, yeah. I wouldn't worry too much about the style changes, this is mostly just OCD-ish behaviour on my part :) We don't really use any kind of formatter, so I'll try to cool that a bit in the future

jokesper:
> so I'll try to cool that a bit in the future
I have no problem with how it is currently done, since it doesn't create any extra work for me.
I just didn't want to create extra work for you.
So feel free to correct things as you have done before.

slotThe:
Thanks!

slotThe:
@jokesper seems like something slipped through: lower-case compose key sequences in the config (e.g., `ü` or `à`) automatically get translated to upper-case with the new default. I haven't investigated this much further, but for now I will revert the commit that changes the default

jokesper:
Hmm I will investigate this further.

jokesper:
The problem is that the `tap` function runs the press followed by the release action. But the new buttons, which were inspired (code wise) by the tap-hold button family which all only assign a press action.
So this results in some wierd entanglement between buttons and concrete keycodes. I.e They can only be truly released via key releases.

PR #869, opened 2024-08-04
Fix broken compose sequences v2

jokesper:
This removes the duplicate sequence for `cent` and the non parseable sequences for `nobreakspace` and `U2008`
(except for command line parameters (what the heck?!))
Also adds some test to check for this.

I hope this doesn't morph into a long running series.
The only missing validation at this point is whether it matches the X11 defined compose sequences
but that would require a lot of effort and using them directly instead of comparing whether we copied them correctly
would also be the better solution as it _would_ also solve #79.


slotThe:
Thanks!

Issue #872, opened 2024-08-12
Misbehaving tap-hold style buttons in tap-macro and other tap-hold style buttons


jokesper:
The `tap-hold` style buttons only have a press action which adds a hook waiting for it's release.
The problem is that said hook only work for keycodes.
This is also a problem for the button in #859.
There are multiple possible solutions:
1. Change hooks to listen to button events instead of key events (possibly requires button ids or checking buttons for equality)
2. Add extra release handler which cancels the hooks registered in the press part (requires unregistering or state)
3. Think up a better solution to describe button which is more robust (a lot of work)

jokesper:
Some parts of the code expect separate press and release actions (`press-only`, `release-only`, `tap-macro`, sub buttons in `tap-hold` style, ...)
While others don't conform to this interface (`tap-hold` style buttons, `around-only`, `around-when-alone`).
So in truth we have different kinds of buttons which are not distinguised.
The perfect solution would probably be to implement some kind of type system and do type checking.
A (hopefully) temporary solution could be adding an optional extra `tap` action which falls back (`around` when using the other variants, only the tap button in the `tap-hold` style).

PR #857, opened 2024-07-07
Cleanup documentation

jokesper:
Added missing documentation and added tests to make sure it stays this way.
This also resulted in `around-next-single` being parsable, after it was broken.
I also removed the option `init` which was added in https://github.com/kmonad/kmonad/commit/dcdf5b5b060af1b6311ef48f3d044cad5f0dcf73 and it's documentation from the code was deleted in https://github.com/kmonad/kmonad/commit/031f865d5a22268072ff1a0cf93e2fa9087ec01f
It was used to specify a program to start when kmonad boots up

PR #873, opened 2024-08-12
Fix tapping in inner

jokesper:
This fixes the problems described in #872 for the following buttons:
- [X] `around-only`
- [x] `around-when-alone`
- [ ] `around-next`
- [ ] `around-next-timeout`
- [ ] `around-next-single`
- [X] `tap-on` (does not exists but has code?!)\*
- [X] `tap-hold`
- [X] `tap-next`
- [X] `tap-hold-next`
- [ ] `before-after-next`
- [X] `tap-next-release`
- [X] `tap-hold-next-release`
- [X] `tap-next-press`
- [X] `multi-tap`
- [X] `tap-macro`
- [X] `tap-macro-release`
- [ ] `layer-next`
- [ ] `sticky-key`
- [ ] `retap` (Not yet implemented. See #882)
- [ ] `stepped-button`

\* `tap-on` is just made more readable. It wasn't broken.

The other unfixed buttons require a deeper rewrite to function on the next button instead of on the next key press (if that is what we want? Probably.).

This also allows us to rerevert 2d11c6306ebd209796474da83efeefd607a30e6c but #859 maybe has other problems with something like `S-(layer-toggle <...>)` which I have seen in old issues. Though there was always another mention of a solution which works better which didn't use that or `(around lsft (layer-toggle <...>))`.

### DIFF
--- a/src/KMonad/Model/Button.hs
+++ b/src/KMonad/Model/Button.hs
@@ -20,6 +20,7 @@ module KMonad.Model.Button
   , HasButton(..)
   , onPress
   , onRelease
+  , onTap
   , mkButton
   , around
   , tapOn
@@ -76,11 +77,15 @@ import qualified RIO.HashSet as S
 -- 'Button' is essentially a collection of 2 different actions, 1 to perform on
 -- 'Press' and another on 'Release'.
 
--- | A 'Button' consists of two 'MonadK' actions, one to take when a press is
+-- | A 'Button' consists of three 'MonadK' actions, one to take when a press is
 -- registered from the OS, and another when a release is registered.
+-- With a third when both should happen in sequence. This will only be used
+-- by other button such as `tap-macro`.
+-- Use 'mkButton' instead of this constructor.
 data Button = Button
   { _pressAction   :: !Action -- ^ Action to take when pressed
   , _releaseAction :: !Action -- ^ Action to take when released
+  , _tapAction     :: !Action -- ^ Action to take when tapped as a sub button
   }
 makeClassy ''Button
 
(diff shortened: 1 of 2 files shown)


src/KMonad/Model/Button.hs
```diff
@@ -81,6 +82,7 @@ import qualified RIO.HashSet as S
 data Button = Button
   { _pressAction   :: !Action -- ^ Action to take when pressed
   , _releaseAction :: !Action -- ^ Action to take when released
+  , _tapAction :: !Action -- ^ Action to take when tapped
```
slotThe, line 85: This honestly feels like a wonky change to me, as what counts as a "tap" or not is not really related to the button, but more to surrounding environment (even things like, say, key repeat)

jokesper: The problem is with buttons which listen for their own release, since they listen to the keycode not the Button.
`tap` in this context referrs to the `tap` function I.e. when you want to press and immideatly release a Button.
The perfect solution would be replacing the detecting of their own release to use the button state. This is a bit difficult since we mix a few things in the affected buttons. If we change the entire hook system to work with buttons instead we have the following problems:
1. ```closure
(defsrc a b)
(deflayer (around-next sft) #(x y))
```
    with `Pa Tb Ra` previously resulting in `Psft Tx Ty Rsft`
    and then resulting in `Psft Tx Rsft Ty` (is this wanted? probably not)
2. `#((around-next sft) a)` binded on `x` and the rest being fallthrough:
    `Tx Ty` priviously `Ta Psft Ty Rsft` and with the change `Psft Ta Rsft Ty` (both are kind of wierd)
3. `layer-next` requires the old system, since want to change the layer prior to button look up.

My point being:
It is not really defined how most buttons should behave in such context. Though something like nested `tap-macro`s have an intended behaviour, it is just broken, since the last key is still held.
This could happen when you have many `tap-macro`s and wanted to reduce work with aliases.
The different `implicit-around` buttons are the prime example for this going wrong. Changing them inside nested contexts is also not the correct solution. Cosider:
`#(A B)` bound on `x` with `fallthrough = true`:
`Px Ty Rx` results in the same problem `implicit-around`s tried to fix: `Psft Ta Rsft Psft Pb Ty Rsft` also written as `A B Y`. since the last button is held.

The affected buttons all have a `press` action but no `release` action. So other buttons who call `tap` priviously `press *> release` now use a third action which defaults to `press *> release` but it can be changed.

T.l.d.r.: I should change the comment

jokesper: Hey @slotThe, I was wondering if there was anything else or if this is ready to merge.
I really don't want to stall #879 as I myself requested a release in #857 two months ago.
Though I also like diligence in code review and respect work and free time.
(Just a ping to assure (mostly to myself) that this is not lost in emails).

slotThe: @jokesper sorry, I've been traveling for work and haven't had much time for this lately. I'll try to get to everything this week.

I still feel like this is more of a bandaid-change, but that's still better than nothing until we properly refactor the keycode handling :)

jokesper:
Ahhh I forgot `tap-hold-next-release` (or maybe pressed undo accidentily)

PR #874, opened 2024-08-12
Fixed CI

jokesper:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


slotThe:
Thanks!

PR #880, opened 2024-09-01
Update compiler and build setup in the dockerfile

lierdakil:
This PR updates the GHC version used in the dockerfile for static builds (which is also used in CI). Also tweaks the CI job, specifically updates actions to latest versions and removes caching for the docker build.

The caching thing is curious. For one, it makes the jobs failure-prone due to "no space left on the device" thing (if you've worked with github actions and docker for some time, you've seen it). Besides, docker image caches don't work like they used to, now that buildx is the default, and the build only takes about 6 minutes from scratch. It's just not worth the trouble at this point.

The old setup for static builds no longer works, but there's now a simpler approach which only requires `-fPIC` and adding `ld-options: -static` to the cabalfile, which the dockerfile patches in-place.

To be honest, I'm fascinated this still works, 8.10 is quite outdated. To be fair, 9.4 is also somewhat outdated, but the root stack.yaml uses lts-21, which is on 9.4, so I've opted to use that for consistency.

P.S. the CI job is run on push and not on PR (by design), so you can't see it succeeding here. But it works as intended on my fork: https://github.com/lierdakil/kmonad/actions/runs/10652049346

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,24 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
-      - uses: actions/cache@v2
-        with:
-          path: image.tar
-          key: ${{ runner.os }}-primes
-
-      - run: |
-          [ -r image.tar ] && docker load -i image.tar || true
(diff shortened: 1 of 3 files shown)


slotThe:
Seems to work fine, thank you! :)

Issue #879, opened 2024-08-31
The current relase is almost a year old


hirak99:
As I was reading the documentation, I noticed it covers new features such as named sources, which do not work in the released version.

Then I realized that the last release is very old - and there has been quite a few new feature additions and bug fixes since then.

The last release 0.4.2 happened on 2023-Oct-07.

Is it time for a new release? :)

jokesper:
I would suggest merging #873 as #859 is partially broken and then doing a release.

slotThe:
Sounds like a good idea to me (although I'm hesitant to go a step further and revert https://github.com/kmonad/kmonad/commit/2d11c6306ebd209796474da83efeefd607a30e6c just yet)

jokesper:
I agree on that though the fixes should still be made.

slotThe:
Done in https://github.com/kmonad/kmonad/releases/tag/0.4.3 :)

PR #882, opened 2024-09-05
Implement retap button

jokesper:
`retap` is a button which allows you to specify another action, which gets executed when this button is retapped.
Personally this is not that useful to me yet since #256 / #524 is not yet fixed / merged ~~and I want to use something like `(tap-hold-next-release 200 a (retap 200  sft (sticky-key 1000  sft)))`.~~\*
I also don't want it to delay #879.
EDIT:
\* Using this with home row buttons was wierd since typing `Button` on dvorak caused `Btton` (`BTton` with #256 fixed) every time. (`u` was my `sft` home mod).

In most other cases something like `multi-tap` is probably preferred.
Since their is currently no use case or request I will convert it to a draft. If anyone still wants it please provide a use case to consider (I don't want to bloat KMonad when not necessary)


jokesper:
Since there hasn't been any interest, I will close it. If anyone wants this in the future, please let me now.

PR #883, opened 2024-09-11
Added names for `rfkill` and `micmute`

jokesper:
Taken from https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h

This leads to a potential new problem:
We already supported the keys, they were just `missing247` and `missing248`
if you did use them in your config you wouldn't get a warning like: "Key has a name", but instead get an error about an unknown key.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 ### Added
 
+- Added names for the keys `rfkill` and `micmute` (#883).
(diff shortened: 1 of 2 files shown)


slotThe:
> This leads to a potential new problem: We already supported the keys, they were just missing247 and missing248 if you did use them in your config you wouldn't get a warning like: "Key has a name", but instead get an error about an unknown key.

I don't think this was documented anywhere; people who went through the code and dug up these names are surely savvy enough to make the required adjustments in this case.

PR #884, opened 2024-09-11
Fixed / changed `tap-next` and `tap-hold-next`

### LINKED ISSUE #572: tap-hold-next takes into account previous key's release
Consider `foo (tap-hold-next 200 x shft)`. If I tap `foo`, but a previous key was released before foo is released, it will emit the 'hold' action rather than the 'tap' action.

e.g. `Pa P@foo Ra 100 R@foo -> a (shift)` (instead of `ax`)

Is this a bug or by design? I was under the impression that `tap-hold-next` only accounts for the *next key* only rather than the *previous key* also.

jokesper:
Fixes #572
Changelog entry is under changed since tutorial specified the old behaviour while quick-reference, comments and the maintainer (in #572) specified it as a bug.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 ### Changed
 
 - `multi-tap` now holds when interrupted by another key while held. (#897)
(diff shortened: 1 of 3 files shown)


jokesper:
`tap-next-press` exists for this purpose

PR #892, opened 2024-09-25
Minor fixes for tutorial.kbd

esuarezsantana:
Fixes the examples that use 'a' instead of 'Ta'.

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -826,11 +826,11 @@
   So, again with the mini-language, we set foo to:
     (tap-hold 200 x lsft) ;; Like tap-next, but with a 200ms timeout
   Then:
-    Tesc            -> x
-    Tesc Ta         -> xa
-    Pesc 300 a      -> A (the moment you press a)
-    Pesc a 300      -> A (after 200 ms)
-    Pesc a 100 Resc -> xa (both happening immediately on Resc)
+    Tesc             -> x
+    Tesc Ta          -> xa
+    Pesc 300 Ta      -> A (the moment you press a)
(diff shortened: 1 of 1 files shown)


keymap/tutorial.kbd
```diff
@@ -826,11 +826,11 @@
   So, again with the mini-language, we set foo to:
     (tap-hold 200 x lsft) ;; Like tap-next, but with a 200ms timeout
   Then:
-    Tesc            -> x
-    Tesc Ta         -> xa
-    Pesc 300 a      -> A (the moment you press a)
-    Pesc a 300      -> A (after 200 ms)
-    Pesc a 100 Resc -> xa (both happening immediately on Resc)
+    Tesc             -> x
+    Tesc Ta          -> xa
```
jokesper, line 830: You could have removed a space in the changed lines and would have gotten a smaller diff.

jokesper (approved):

keymap/tutorial.kbd
esuarezsantana: These spaces were added to preserve aligment of the arrows in this block of examples. It looks better to me. Is it relevant the size of the diff?

jokesper: In general diff size is not that relevant just a good boni.
What I meant is that you could have removed a space in the lines you did edit since you would keep alignment but also have a small diff.
Though just my inner perfectionist speaking.
Adding the extra work of ammending the commit and force pushing, is simply not necessary.

slotThe:
I agree that `Pesc Ta 100 Resc->` would look a bit weird, just to make the diff smaller (although we are using this for that reason in several portions of the code :))

Thanks!

PR #894, opened 2024-09-27
Added keycodes from windows and merge tables

### LINKED ISSUE #785: Cannot translate to windows keycode
Posted this in the discussions section but realized this may be a better place. I run in a Windows environment and built KMonad from top of tree. 50% of the time I start up KMonad, I receive the following message.

```
Encountered error in KeySink: Cannot translate to windows keycode: Missing254
Encountered error in <emitter_proc>: Cannot translate to windows keycode: Missing254
```

What does this message mean, and how can I debug it? Thanks.

jokesper:
Fixes #785.

The tables have been merged to reduce chance of inconsistencies. This was possible due to the following observations:
1. `winCodeToKeyCode` always chooses the keycode with lowest value when duplicates exist.
2. `keyCodeToWinCode` always chooses the highest resulting possible keycode.

(windows keycode `0xFE` as delete would change this, though it does not exist in the KMonad world (and probably never will).)


kmonad.cabal
```diff
@@ -40,6 +40,7 @@ library
     -Wno-unused-imports
   build-depends:
       base >= 4.12 && < 5
+    , foldable1-classes-compat
```
slotThe, line 43: This isn't already in our transitive dependencies, right? I'm sort of wary when it comes to introducing a dependency for one function (or two, in this case, I suppose)

jokesper: `foldable1-classes-compat` is a compatibility package. It's only needed since we allow base <= 4.18.0.0 (GHC 9.6.1).
I didn't want to remove old compiler support, since they aren't that outdated

slotThe: Sure, but the alternative is just to use the "ordinary" `minimumBy` and `maximumBy` from `Data.Foldable`, no? Since the lists are static, I a priori see no reason to prefer the `Foldable1` definition

jokesper: Technically yes, but I personally prefer the method where the compiler guarantees, that all cases are handled. Especially, since this function is already in base, we just use an outdated version (lts-22 would be enough).
That being said, it's not my project, and I do prefer _a_ good solution now, over _my_ far later / never.

slotThe: I think it's generally a good idea to keep supporting whatever GHC version the current Debian stable is running, which AFAIR is 9.0 right now. This should enable everyone to build the project with their system GHC (just using cabal)

PR #896, opened 2024-09-27
Refactor with minor improvements

jokesper:
While taking a look around some code, I found some areas which could be somewhat improved to aid readability.

1. Reduction in binary size (-split-sections)
    (What happened between 0.4.2 and 0.4.3 that the code size increased from 3.5MiB to 13.8MiB?).
2. Reenable ci for windows (We had this the whole time?!).
3. Interesting C code, which got improved (slightly) to remove warnings in Windows CI.
4. #894 would have fit perfectly here.
5. Syntax change to be consistent and have clearer diffs when changing CI.
6. More windows CI + some code cleanup.
7. Why did we ever ignore warnings?
8. Make use of a Custom Prelude which is a Prelude and doesn't have to be imported.
9. Why do we have a directory with only one file in it. (app/)

t.l.d.r.: No major changes just some minor ones which bothered me while working on the code base.


jokesper:
Hmm, the Windows build using Wine sometimes errors with `Cabal-7125` which appears to indicate segfaulting of cabal.
Is this a wine bug or cabal? Either way nothing to fix on our side except maybe not building haskell through wine.

PR #898, opened 2024-09-27
Fix tapping for `tap-hold-next-release`

jokesper:
I forgot to actually fix `tap-hold-next-release` while making #873.


slotThe:
Thanks!

PR #906, opened 2024-10-07
Fix tapping for `tapMacro` and `tapMacroRelease`

### LINKED ISSUE #904: Multi-tap behaviour changed
MI2 #(a (multi-tap 700 XX #(bspc b (multi-tap 700 XX #(bspc c)))))

Defining aliases like this in order to recreate the behaviour of old phone keyboards used to work in 0.4.2 but in 0.4.3 it prints the first letter no matter how many times the key is pressed. Maybe I was exploiting some kind of bug that is now fixed

jokesper:
Fixes #904.

I accidentily switched `True` and `False` in #873. We should really add tests to check for button behaviour, which is sadly a bit hard, if we don't want a test to require an IO action (heavy usage of `TMVar`s) (which it does anyway, since Spec wraps an IO monad) or want to have two distinct implementations of `MonadK`.

Sorry, I hope this was the last bug I introduced (at least in #873).

We could probably also do prereleases, which of course only makes sense if releases are more common.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 ### Fixed
 
+- Fixed `tapMacro` and `tapMacroRelease` behaviour which was slightly broken in #873 (#906)
+
 ## 0.4.3 – 2024-09-11
 
 ### Added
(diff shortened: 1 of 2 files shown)


slotThe:
Thanks. Perhaps I should have tested #873 a bit more thoroughly…

Issue #867, opened 2024-07-31
please implement keyrepeat inside kmonad


devvythelopper:
The issue is simple: keyrepeat under x11 works fine for normal keys, but not for composed keys (keys composed by kmonad itself). If kmonad implemented keyrepeat directly, then composed keys could be repeated as well.

PR #913, opened 2024-10-27
Fix keyrepeat not working in the tty

### LINKED ISSUE #188: Key repeat in the tty?
Hello,
I wanted to use kmonad to remap the backspace and caps lock keys, I used it since it can work with low-level keyboard input which is suitable for use with tty (unlike something like xmodmap). However, there is no key repeat or at least an option that makes it possible I know of, I think key repeat works in X server due to other X specific programs, I think this may be related to #82 or #184 but for linux.
Thanks for reading my issue.

### LINKED ISSUE #543: Hold key doesn't work as expected in tty
Hello. When i hold `h` key in my vim or file manager I expect scrolling to bottom.
With enabled kmonad it works as expected in `sway` wm but it doesnt in tty. It does only one jump and stop.

How to fix it?

jokesper:
Apparently we just needed to set an extra flag.
See [Blagus's comment](https://forums.raspberrypi.com/viewtopic.php?t=45041#p1066932) linking to [Using uinput driver in Linux-2.6.x](https://www.geocities.ws/gnumohan/Linux/uinput.txt).

Fixes #188. Though we should really be looking into implementing it in KMonad like #867 suggest.
Also fixes #543.

### DIFF
--- a/c_src/keyio.c
+++ b/c_src/keyio.c
@@ -21,6 +21,8 @@ int acquire_uinput_keysink(int fd, char *name, int vendor, int product, int vers
   for (i=0; i < 256; i++) {
     ioctl(fd, UI_SET_KEYBIT, i);
   }
+  // Enable key repeat for output keyboard
+  ioctl(fd, UI_SET_EVBIT, EV_REP);
 
   // Set the vendor details
   struct uinput_setup usetup;
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 - Fixed `tapMacro` and `tapMacroRelease` behaviour which was slightly broken in #873 (#906)
 - Fixed keycode translation problem on windows (#894)
+- Fixed keyrepeat not working in tty on linux (#913)
 
 ## 0.4.3 – 2024-09-11
 


c_src/keyio.c
```diff
@@ -21,6 +21,8 @@ int acquire_uinput_keysink(int fd, char *name, int vendor, int product, int vers
   for (i=0; i < 256; i++) {
     ioctl(fd, UI_SET_KEYBIT, i);
   }
+  // Enable key repeat for output keyboard
+  ioctl(fd, UI_SET_EVBIT, EV_REP);
```
jian-lin, line 25: I am not sure why we do not check `ioctl` return values in this `acquire_uinput_keysink` function.  Probably we should do that like what we do in other functions.

It may be out of the scope of this PR thought and we can do that in a future PR.

jokesper: Probably, though I think checking the length by the `strcpy` is also something important, since you can trivially type more than 80 characters in the `uinput-sink <name>` parameter.

jokesper: I fixed the overflow and some types in [my local branch](https://github.com/jokesper/kmonad/tree/fix-c). Though this doesn't add checks in the C code. If you want you could do it. (Either independently or as a PR against my branch)

slotThe:
Thanks! I agree that the C code could probably use some more love, in case anyone wants to do that

PR #914, opened 2024-10-27
#524 rebased on master for easier merging (and testing)

jokesper:
I "rebased" #524, for easier testing and merging.
I had to squash the commits, because they contained merges with conflicts, which I didn't want to reresolve.
Thanks jchtt for this amazing PR!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 - Fixed `tapMacro` and `tapMacroRelease` behaviour which was slightly broken in #873 (#906)
 - Fixed keycode translation problem on windows (#894)
+- Fixed problems with tap-next-release and similar (see #256) (#524, #914)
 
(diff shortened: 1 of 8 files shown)


jokesper:
I noticed that this PR can lead to unintended key repeat.

E.g.: when using KMonad in the tty via #913 we could have problems with `hold` since the delay is 250ms by default.
This is mostly a Problem with the more complex buttons, and only happened accidentily (for me) with #916.

The trivial fix is setting the keyrepeat delay higher. The best fix would probably be #867.
Though any solution to #867 would probably break games somewhat.

jokesper:
I think this greatly decreases readabilty of logs. Maybe we should fix this first

jokesper:
Hey @fleutot, since you are already running this PR, I wanted to inform you, that the last force-push did have some changes (therefore the extra commit also). I applied the fix of this PR to the new button `tap-hold-next-press` and to the newly added holding behaviour of `multiTap`.

So depending on whether you use any of them you may want to redo the merging.

PR #927, opened 2024-11-23
Small update for macOS installation.md

SanderHestvik:
Make it easier to see that kext and dext are two different paths


slotThe:
Thank you for the PR! However, I feel like the current code is clear enough, personally, and don't really see the need to add raw raw HTML into markdown files for things like this

jokesper:
If raw HTML is the problem, a newline more would have probably done something similar

PR #935, opened 2024-12-20
Added more keys from `input-event-codes.h`

### LINKED ISSUE #912: Add support for macro keys
The linux kernel defines macro keys

FROM input-event-codes.h 

#define KEY_MACRO1			0x290
#define KEY_MACRO2			0x291
#define KEY_MACRO3			0x292
#define KEY_MACRO4			0x293
#define KEY_MACRO5			0x294
… continues up to 30

would it be simple to add some of them in the Kmonad/Keyboard/Keycode.hs file so that we can map to those keys?

### LINKED ISSUE #933: Fn is unusable (again) with kmonad
Hi. I am on linux, using macbook air keyboard.

<details>
<summary>My config.kbd</summary>

```
(defcfg
	input	(device-file "/dev/input/by-path/platform-23510c000.spi-cs-0-event-kbd")
	output	(uinput-sink "KMonad kbd")
	fallthrough true
)

(defsrc
	esc	f1	f2	f3	f4	f5	f6	f7	f8	f9	f10	f11	f12
	grv	1	2	3	4	5	6	7	8	9	0	-	=	bspc
	tab	q	w	e	r	t	y	u	i	o	p	[	]	ret
	caps	a	s	d	f	g	h	j	k	l	;	'	\
	lsft	rctrl	z	x	c	v	b	n	m	,	.	/	rsft
	fn	lctl	lalt	lmet			spc			rmet	ralt
)

(defalias
	sxtr	(tap-next caps (layer-toggle extra))
	fn	(cmd-button "hyprctl dispatch exec \"tofi-drun | xargs hyprctl dispatch exec --\"")
)

(deflayer default
	esc	f1	f2	f3	f4	f5	f6	f7	f8	f9	f10	f11	f12
	grv	1	2	3	4	5	6	7	8	9	0	-	=	bspc
	tab	q	w	e	r	t	y	u	i	o	p	[	]	ret
	@sxtr	a	s	d	f	g	h	j	k	l	;	'	\
	lsft	rctrl	z	x	c	v	b	n	m	,	.	/	rsft
	fn	lctl	lalt	lmet			spc			rmet	ralt
)

(deflayer extra
	esc	f1	f2	f3	f4	f5	f6	f7	f8	f9	f10	f11	f12
	grv	1	2	3	4	5	6	7	8	9	0	-	=	bspc
	tab	1	2	3	4	5	6	7	esc	9	0	[	]	ret
	caps	ä	ä	_	_	_	_	left	down	up	right	'	\
	lsft	ä	!	@	#	$	%	^	&	\(	\)	/	rsft
	fn	lctl	lalt	lmet			spc			rmet	ralt
)


```

</details>

Wev with kmonad off:
```
[14:     wl_keyboard] key: serial: 47259; time: 7175272; key: 472; state: 1 (pressed)
                      sym: XF86Fn       (268964304), utf8: ''
[14:     wl_keyboard] key: serial: 47260; time: 7175393; key: 472; state: 0 (released)
                      sym: XF86Fn       (268964304), utf8: ''
```

But with kmonad running, fn button is not emitted.

Additional info:
I was able to use kmonad and bind something to fn using patch I made from https://github.com/kmonad/kmonad/pull/475/files. I did not upgrade quite a while and now I see kmonad is finally getting updates. So what is the correct way to handle this? And yes, I know about fn question in faq.

Also you can see `fn` alias for my launcher, it does not work either, though, now I am using hyprland key bindings, but i'll leave it as an option here too.

jokesper:
This fixes #933 and also fixes #912.

I have mostly generated them automatically via
<details><summary>code golfing</summary>

```haskell
#!/usr/bin/env runhaskell
{-# LANGUAGE ViewPatterns #-}

import Control.Arrow
import Control.Monad
import Data.Bool
import Data.Char
import Data.Foldable
import Data.Foldable1 (foldMap1)
import Data.Function
import Data.Functor
import Data.List.NonEmpty (NonEmpty (..), groupAllWith)
import Data.List.NonEmpty qualified as N
import Data.Tuple
import Numeric
import RIO.List

isNum [] = False
isNum ('0' : 'x' : hex) = all (`elem` ['0' .. '9'] ++ ['a' .. 'f']) hex
isNum ds = all (`elem` ['0' .. '9']) ds

formatName :: String -> String
formatName = filter (/= '_') . (zipWith (bool toLower toUpper . (== '_')) =<< ('_' :))

parse' lines =
  let
    [(read @Int -> max) : _] = words <$> (toList . stripPrefix "#define KEY_MAX" =<< lines)
    keys =
      map (foldMap1 fst &&& snd . N.head)
        . groupAllWith snd -- aliases defined by duplicate number
        . map
          ( \((name, read @Int -> code) :| (map snd -> aliases)) ->
              (formatName <$> N.reverse (name :| aliases), code)
          )
        . groupAllWith fst -- aliases defined by name
        . map ((bool swap id =<< isNum . snd) . \[k, v] -> (k, v))
        $ extract lines
    missings' = map (first (:| [])) . filter (not . (`elem` map snd keys) . snd) $ missings max
   in
    sortOn snd $ keys ++ missings'

missings max =
  [0 .. max - 1] <&> \n ->
    ("Missing" ++ if n < 256 then show n else "0x" ++ showHex n "", n)
extract =
  toList
    . stripPrefix ["#define"]
    . take 3
    . words
    <=< takeWhile (not . isPrefixOf "#define KEY_MAX")
      . dropWhile (not . isPrefixOf "#define KEY_RESERVED")

main = do
  file <- lines <$> readFile "/usr/include/linux/input-event-codes.h"

  let keys = parse' file

  let enum = join . ("  = " :) . intersperse "\n  | " $ N.head . fst <$> keys
  let aliases = fst <$> keys
  let aliases' =
        join . ("  , " :) . intersperse "\n  , " $
          filter (not . null . N.tail) aliases <&> \(key :| aliases) ->
            fold
              [ "("
              , key
              , ","
              , replicate (21 - length key - 2) ' '
              , "["
              , join . intersperse ", " $ aliases <&> \a -> '"' : (toLower <$> dropPrefix "Key" a) ++ "\""
              , "])"
              ]

  putStrLn enum

  putStrLn aliases'
```
</details>

Sometimes I had to capitalize letters, since they were seperated wierdly in `input-event-codes.h`.

This not only includes keys but also buttons (e.g. on the mouse). I choose to include them and not actively support them further rather then to add more `missing`s

This also significally increases compile time for the `Keycodes.hs` file, though I don't know how to fix this


src/KMonad/Keyboard/Keycode.hs
```diff
@@ -301,8 +301,525 @@ data Keycode
...
+  | KeyRed
+  | KeyGreen
+  | KeyYellow
+  | KeyBlue
+  | KeyChannelUp
+  | KeyChannelDown
+  | KeyFirst
+  | KeyLast
+  | KeyAb
+  | KeyNext
```
jokesper, line 461: This clashes with the alias for `KeyNextSong`, how should we remedy this?
(caught by tests in #940 (merged locally))

jokesper: I choose to use the HID names instead for `KeyNext` and `KeyPrevious`. This should maybe also be done for `KeyFirst`, ....

I think the best option is to change to HID names for internal use as a breaking change, though this is not really urgent.

slotThe:
Thanks!

> This also significally increases compile time for the `Keycodes.hs` file, though I don't know how to fix this

Beyond abandoning the big sum type for the keys it probably is what it is. This type of making illegal states unrepresentable is quite alluring, although now it's *really* big :) A nucleation site for future thoughts, perhaps

jokesper:
> This also significally increases compile time for the `Keycodes.hs` file, though I don't know how to fix this

After some testing with the instances, I tested the following instance combinations:

```
Eq                            1.07
Eq, Data                      1.81
Eq, Generic                  10.35
Eq, Data, Generic            11.26
Eq, Generic, Hashable        11.07
Eq, Data, Generic, Hashable  12.17
```

It looks like the `Generic` instance takes about 9 extra seconds.

A simple test of splitting the data type into 3 sub types results in 6.8 seconds from scratch. Though the simplest solution would probably be a handwritten hashable function.

PR #937, opened 2024-12-26
[macOS] Upgrade Karabiner-DriverKit-VirtualHIDDevice to v5.0.0

homovetus:
This PR upgrades Karabiner-DriverKit-VirtualHIDDevice to version v5.0.0.

## Improvements
Starting with version **v4.0.0**, Karabiner-DriverKit ensures binary compatibility for client applications. As a result, kmonad no longer need to be recompiled when upgrading Karabiner-DriverKit.

### macOS Compatibility:
Support for macOS 12 and earlier versions has been discontinued.

### Daemon Management:
- As of v5.0.0, Karabiner-DriverKit requires the daemon to be explicitly managed by the user.  
- The installation now includes:
  - A requirement to start the daemon for dext.
  - Link to the plist file for autostart daemon using `launchd`.

### DIFF
--- a/c_src/mac/dext.cpp
+++ b/c_src/mac/dext.cpp
@@ -23,10 +23,13 @@ int init_sink() {
     /**/
     client->connected.connect([copy] {
                                   std::cout << "connected" << std::endl;
-                                  copy->async_virtual_hid_keyboard_initialize(pqrs::hid::country_code::us);
+                                  pqrs::karabiner::driverkit::virtual_hid_device_service::virtual_hid_keyboard_parameters parameters;
+                                  parameters.set_country_code(pqrs::hid::country_code::us);
+                                  copy->async_virtual_hid_keyboard_initialize(parameters);
                               });
     client->connect_failed.connect([](auto&& error_code) {
                                        std::cout << "connect_failed " << error_code << std::endl;
+                                       std::cout << "You may need to start Karabiner" << std::endl;
                                    });
     client->closed.connect([] {
                                std::cout << "closed" << std::endl;
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 ### Added
 
 ### Changed
+- Update Karabiner-DriverKit to 5.0.0 (#937)
+  You will now need to [start the server yourself](doc/installation.md#starting-the-dext-daemon).
 
 ### Fixed
 
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -338,8 +338,8 @@ Simply run these commands in Windows PowerShell:
 
 KMonad supports macOS 10.12 to 10.15 (Sierra, High Sierra, Mojave, and
 Catalina) and macOS 11.0 (Big Sur). When using driverkit-based extension
-v2.1.0 and later, KMonad also supports macOS 13.0 (Ventura) and 14.0
-(Sonoma).
+v5.0.0 and later, KMonad also supports macOS 13.0 (Ventura), 14.0
+(Sonoma) and 15.0 (Sequoia).
 
 Note: Under macOS, `kmonad` uses one of two "system extensions" to
 post modified key events to the OS. For macOS Catalina and prior, we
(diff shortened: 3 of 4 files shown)


c_src/mac/keyio_mac.hpp
```diff
@@ -8,12 +8,6 @@
 #include <mach/mach_error.h>
 #include <AvailabilityMacros.h>
 
-/* The name was changed from "Master" to "Main" in Apple SDK 12.0 (Monterey) */
-/* MAC_OS_X_VERSION_12_0 does not exist :( */
-#if MAC_OS_X_VERSION_MAX_ALLOWED < 120000
-    #define kIOMainPortDefault kIOMasterPortDefault
-#endif
-
```
jokesper, line 16: Why?!

This file is common across dext and kext. This would break every user of kext, without completly removing support. I would suggest simply dropping this deletion

c_src/mac/dext.cpp
```diff
@@ -23,7 +23,9 @@ int init_sink() {
     /**/
     client->connected.connect([copy] {
                                   std::cout << "connected" << std::endl;
-                                  copy->async_virtual_hid_keyboard_initialize(pqrs::hid::country_code::us);
+                                  pqrs::karabiner::driverkit::virtual_hid_device_service::virtual_hid_keyboard_parameters parameters;
+                                  parameters.set_country_code(pqrs::hid::country_code::us);
+                                  copy->async_virtual_hid_keyboard_initialize(parameters);
                               });
     client->connect_failed.connect([](auto&& error_code) {
                                        std::cout << "connect_failed " << error_code << std::endl;
```
jokesper, line 31: Could you also add something like:
```diff
+                                       std::cout << "You may need to start Karabiner " << std::endl;
```
since it isn't started automatically anymore and this would be a very bad error message then.

jokesper (dismissed):
I tried to do this myself a while back but for some reason KMonad refused to work on mac (that's why I tried to update karabiner in the first place), though since I couldn't confirm it worked, I didn't open a PR. Since we have the same / similar / you have slightly better changes (except for dropping the ifdef in `keyio_mac.hpp`), I would give my stamp of approval ~~(though unnecessary, since I'm not a maintainer)~~.
Great work!

jokesper:
I have applied my requested changes and would say this is ready to merge now.
Except since I cannot test it, I would like to await testing of other users.
Since this is the same thing blocking #980, maybe you and @jsimonetti could test the others changes.

Alternatively I maybe @thoelze1 could do it, since he contributed a lot of the macos parts (I think)

jsimonetti:
I will test this weekend. 👍

jsimonetti:
So, I finally managed to test this PR and it works as expected.

I did have trouble with the old Karabiner VirtualHIDDevice driver installation and had to completely remove the old version, reboot and install the new version.
During this process I also somehow managed to run the old kmonad binary with the new driver. This ended up in a broken system, as keys where captured, but not injected back in.
For this reason alone I would suggest cutting a new release before (and after) merging this PR, as this is NOT backwards compatible.

jokesper:
Thank you for the feedback.
I think merging your PR and maybe #982 followed by a release would probably be the right action.

jokesper:
I just remembered:
Since we don't provide prebuild macos binaries, the common install method would be cloning and building from source. I don't think many go through the trouble of checking out the release tag.
So waiting with this till after the release would affect only nix users with upstream nixpkgs.
Are there any other package managers which package kmonad on macos?

jsimonetti:
> Are there any other package managers which package kmonad on macos?

Perhaps homebrew? I can't find it on brew.sh though. I use nix/home-manager myself so can only speak about that.

jokesper:
Would it make any difference for `home-manager`? It should only be a breaking change relating to installation and service starting. `home-manager` should take care of both shouldn't it?

jsimonetti:
> Would it make any difference for `home-manager`? It should only be a breaking change relating to installation and service starting.

That is true. Things will only go bad if you follow `master` and don't lock a release tag.

(just to clarify, i use home-manager, but I still have to 'manually' create a module for the various launchd scripts, etc)

jsimonetti:
Is there anything left blocking this?

jokesper:
I think making a release then merging is the best approach.
I'm currently missing permissions to create a release on hackage, so I asked @slotThe.
He wanted to do it last weekend but evidently it didn't happen, so I will ask him again.

slotThe:
> I think making a release then merging is the best approach. I'm currently missing permissions to create a release on hackage, so I asked @slotThe. He wanted to do it last weekend but evidently it didn't happen, so I will ask him again.

Whoops :)

This week, promise!

slotThe:
Thanks!

PR #940, opened 2024-12-29
Deduplicate compose sequence and added more tests

jokesper:
Added a test for keycode duplicate naming and removed unnecessary compose sequence (defined twice).


slotThe:
Thanks! Sorry for not keeping up with things lately, I've been super busy with life and other things. Hoping to get back into the swing of things in the new year :)

PR #941, opened 2024-12-30
Added option to enable KMonad in the initrd for NixOS

jokesper:
While I'm at it, this also fixes the descriptions of the options.

Furthermore we change the `cmp-seq-delay` default, to the one from KMonad.

Though a default of 0 is a bit wierd, since something like 5 has no real drawback but fixes issues. Personally I would be in favor of increasing it (from 0) or even better use #908 instead.

Will make a PR for it, but github cannot do dependent PRs, so this will wait till #908 is merged.

Sorry for mixing multiple things in the same PR

### DIFF
--- a/nix/nixos-module.nix
+++ b/nix/nixos-module.nix
@@ -167,10 +167,7 @@ in
       isSystemUser = true;
     };
 
-    services.udev.extraRules = ''
-      # KMonad user access to /dev/uinput
-      KERNEL=="uinput", MODE="0660", GROUP="uinput", OPTIONS+="static_node=uinput"
-    '';
+    hardware.uinput.enable = true;
 
     systemd.paths =
       builtins.listToAttrs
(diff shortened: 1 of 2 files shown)


slotThe:
I think a certain split is unavoidable, since nixpkgs now has its own kmonad module, and ours has existed long enough that dropping support is not really an option

nix/nixos-module.nix
```diff
@@ -161,29 +170,52 @@ in
...
+      users.users.kmonad = {
+        description = "KMonad system user";
+        group = "kmonad";
+        isSystemUser = true;
+      };
 
-    services.udev.extraRules = ''
-      # KMonad user access to /dev/uinput
-      KERNEL=="uinput", MODE="0660", GROUP="uinput", OPTIONS+="static_node=uinput"
-    '';
```
slotThe, line 179: This is a bit mangled in these changes; I think we should perhaps split it out into a separate one?

slotThe:
Thanks!

PR #949, opened 2025-01-04
Rework changelog to use git trailers

jokesper:
The current changelog system as used by others is to ignore `changelog.md` entirely.
This leads to a lot of work when making a new release.

The changelog system documented in `CONTRIBUTING.md` leads to lots of mergeconflicts.
But has the benefit of reducing the work when making releases.

Instead of those systems, I propose to use git trailers and automatically extract the changelog from them.

Transitioning will require some manual work:
1. Open pull requests need to be handled (I.e. either they add the changelog entry, or it has to be done after merging)
    Currently only #937 is missing one from the newer ones.
2. The first release needs to merge the `Unreleased` section with the generated entries.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,8 @@
+### Description
+
+Include a description for your changes, including the motivation behind them.
+
+### Checklist
+
+- [ ] I've read [CONTRIBUTING.md](https://github.com/kmonad/kmonad/blob/master/CONTRIBUTING.md)
+- [ ] I added a changelog entry as described in [Add a Changelog entry](https://github.com/kmonad/kmonad/blob/master/CONTRIBUTING.md#add-a-changelog-entry)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,25 @@ request:
 - A brief summary of how your commits fit together to achieve this (if
   necessary).
 
-Please also remember to update the `CHANGELOG.md` file, as well as any other
-documentation that your pull request touches upon.  For example, when
-implementing a new feature, you should update the `tutorial.kbd` file
-accordingly.
+Please also remember to update any documentation that your pull request touches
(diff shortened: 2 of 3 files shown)


jokesper:
> perhaps just having a pull request template that reminds people of adding user-facing changes to the changelog is enough

maybe, though there are two problems at play here
1. documenting changes
2. doing it in a way, that doesn't cause much overhead.

At first glance I would agree with the changelog file, though it has the problem of causing conflicts if two prs change something.

I.e.:
1. PR 1 may fix a bug and add it to the bottom of the changelog.
2. PR 2 fixes a different bug and adds it to the bottom of the changelog.
3. PR 1 gets merged
4. PR 2 has a merge conflict, since both changed the same area of "code" (the bottom of the changelog)

This problem increases for every PR having changes in the same area of the changelog.

Possible solutions would be:
1. Not changing the changelog in the PR, i.e. a maintainer could add it after merging /  when making a release.
     This slightly increases the burden on the maintainer to find a short yet descriptive wording.
2. taking the PR title or commit title as a changelog. How would one differentiate between userfacing changes and internal changes. What if something about a feature is changed but the feature is new in this release (this is also a problem if it's changelog entry would have had to be changed)
3. taking something from the commit body. This is what trailers are.

Of all of these adding a line to the commit body, isn't that big of a barrier to entry I think.
You already have to write a commit message and adding a body isn't any different.

Maybe we should change the suggested template, to leave out the more complicated concepts and just give an example.

slotThe:
> At first glance I would agree with the changelog file, though it has the problem of causing conflicts if two prs change something.

Honestly, in those cases I usually just fix the conflict myself and then merge the PR.

Plus, I like having a user-facing changelog that people can just go to to see what changed (without having to clone the repo first and executing a script). If we wanted to keep that via some other automation, then we're right back at "how to handle merge conflicts".

I'd say for now let's just provide a proper PR template to remind people how to do this, and if it ever really becomes a pain point we can still think about using git trailers to generate a changelog instead.

PR #959, opened 2025-01-25
Simplify installation

### LINKED ISSUE #932: how to i download a binary for windows?
when tunning the binary from https://github.com/kmonad/kmonad/releases/tag/0.4.3 and runing it, i get an error message
image

jokesper:
Cherry-picked from #896. With some squashing.

Simplify installation by:
- reducing the binary size
- providing a docker image for windows.
  since this is designed to run on linux it's help is limited for windows users.
  It's real purpose is to allow generation of windows release binaries as often requested.
  (fixes #932)
- Provide flags for compilation on macos in `cabal` file.

The windows dockerfile sometimes errors while building a random package.
Hopefully ghc will become runtime retargetable soon.
Alternatively we could build our own cross compiler, but that would greatly increase compile time.
Furthermore the `TemplateHaskell` code would cause some problems.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,15 +7,27 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - linux
+          - windows
     steps:
 
-      - uses: actions/checkout@v4
+      - name: Clone project
+        uses: actions/checkout@v4
 
-      - run: docker build . -t kmonad-builder
+      - name: Build release
+        run: docker build -f ci/Dockerfile.${{ matrix.target }} . -t kmonad-builder
 
-      - run: docker run --rm -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
+      - name: Save binary
+        run: |
+          mkdir output
+          docker run --rm -v ${PWD}/output:/host/ kmonad-builder bash -c 'cp -vp /output/* /host/'
 
-      - uses: actions/upload-artifact@v4
+      - name: Upload binary
+        uses: actions/upload-artifact@v4
         with:
-          name: kmonad-binary
-          path: kmonad
+          name: kmonad-${{ matrix.target }}
+          path: 'output/*'
--- a/ci/Dockerfile.linux
+++ b/ci/Dockerfile.linux
@@ -8,7 +8,7 @@ RUN stack update
 COPY stack.yaml kmonad.cabal ./
 # We edit the `stack.yaml` file instead of passing `--ghc-options`,
 # since we want to apply those options to all packages. Not just KMonad.
-RUN sed -i '/ghc-options/ a\  $everything: -fPIC -split-sections' stack.yaml
+RUN sed -i '/ghc-options/ a\  $everything: -split-sections' stack.yaml
 RUN sed -i '/executable kmonad/ a\  ld-options: -static' kmonad.cabal
 RUN stack --no-install-ghc --system-ghc --skip-ghc-check -j8 build --only-dependencies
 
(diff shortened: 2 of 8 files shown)


slotThe (commented):
Looks fine to me, thanks! Cc'ing @lierdakil for the (Linux-specific—I wouldn't dare to inflict Windows on anyone) docker related changes, if you want to take a look

ci/Dockerfile.linux
```diff
@@ -0,0 +1,15 @@
+# syntax=docker/dockerfile:1.7-labs
+FROM lierdakil/alpine-haskell:9.4.8
+
+WORKDIR /usr/src/kmonad/
+RUN apk --no-cache add git
+RUN stack update
+
+COPY stack.yaml kmonad.cabal ./
+RUN sed -i '/ghc-options/ a\  $everything: -fPIC -split-sections' stack.yaml
```
lierdakil, line 9: I find this patching of `stack.yaml` to be unnecessarily brittle. Is there a strong reason to prefer it over specifying `ghc-options` on the command line?

jokesper: yes, we cannot specify the `$everything` part. having a `ghc-options` on the command line only applies it to the current package. This was the original problem I noticed. I will make a comment in the code to clarify it.

.github/workflows/docker.yml
```diff
@@ -7,15 +7,27 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true # Use "Stack CI" for all errors instead.
```
lierdakil, line 11: I'd rather not set fail-fast here (or rather set it to false, as it's true by default IIRC). This basically says "if windows build fails, cancel linux build and vice versa". But those are quite different from one another, and there are many places where either could fail independently. So IMO it's more useful to always have both run to completion.

ci/Dockerfile.linux
jokesper: I don't quite now h relevant the effects of `-fPIC` is. I included it because in the past both `-split-sections` and `-fPIC` where applied to every package and it got broken while the dockerfile was simplified in #880.

jokesper: Do we even need it at all? We don't build a shared library, so we don't really need position independent code do we?

lierdakil: So without `-fPIC` static build failed, if memory serves, because of some GHC quirk (the cabalfile does have a library component after all). It's possible that this is not the case anymore and `-fPIC` could be dropped, not sure.

lierdakil (approved):
LGTM.

jokesper:
I will try to run the dockerfile without `-fPIC`. If it builds, I will remove it. If not
I will try again with it only for `kmonad`.

Thanks for reviewing

ci/Dockerfile.linux
jokesper: It seems it works perfectly fine without. The binary is works also.

jokesper:
I think I found the problem with the windows build (kind of):

My last four error messages where the following:

1.
    > 38.37 ghc-pkg-9.6.6.exe: C:\users\root\AppData\Local\cabal\store\ghc-9.6.6\package.db\package.cache: you don't have permission to modify this file

2.
    > 63.96 ghc-pkg dump failed: Z:\haskell\bin\ghc-pkg-9.6.6.exe: createProcess: invalid argument (Invalid argument)

3.
    > 224.4 <no location info>: error:
    > 224.4     Z:\\haskell\\lib\\../mingw/bin/clang.exe: builderMainLoop: invalid argument (Invalid argument)

4.
    > 104.9 ghc-pkg-9.6.6.exe: C:\users\root\AppData\Local\cabal\store\ghc-9.6.6\package.db\package.cache: you don't have permission to modify this file
    
This seems to indicate, it's some kind of locking problem with `ghc-pkg`, though I don't know what `clang` is doing there.
The relevant issue is probably [#13194 of GHC](https://gitlab.haskell.org/ghc/ghc/-/issues/13194). Maybe it's a bug in wine, resulting in some assumption `ghc-pkg` makes being incorrect. I will try without the `-j8`, though that will significantly increase build times (I think).

PR #963, opened 2025-02-02
Fix NixOS install instructions

### LINKED ISSUE #961: Nix Flake broken
After a recent update I am getting an error while attempting to install/update the Kmonad flake:

```
fatal: https://github.com/kmonad/kmonad?dir=nix/info/refs not valid: is this a git repository?
warning: could not read HEAD ref from repo at 'https://github.com/kmonad/kmonad?dir=nix', using 'master'
nix-output-monitor error: ParseNixJSONMessageError "InternalException (DocumentError {path = \"/type\", errorMsg = \"invalid activity result type: 112\"})" "{\"action\":\"start\",\"id\":697373839851522,\"level\":4,\"parent\":0,\"text\":\"fetching Git repository 'https://github.com/kmonad/kmonad?dir=nix'\",\"type\":112}"
fatal: https://github.com/kmonad/kmonad?dir=nix/info/refs not valid: is this a git repository?
```

jokesper:
We simplify the install url, since `submodules` seems broken and is unneeded anyway.

Fixes #961.

Btw. was `submodules` every documented anywhere? All I could find is random discussion threads.
No actual documentation.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -121,7 +121,7 @@ instead of a manual configuration.
 
    ``` nix
    kmonad = {
-     url = "git+https://github.com/kmonad/kmonad?submodules=1&dir=nix";
+     url = "github:kmonad/kmonad?dir=nix";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    ```


slotThe:
> We simplify the install url, since `submodules` seems broken and is unneeded anyway.

I don't think it was `submodules`, but rather the URL format (which seems like a bug in nix, but since I still can't reproduce this I also can't report it anywhere).

> Btw. was `submodules` every documented anywhere? All I could find is random discussion threads. No actual documentation.

It's for the MacOS stuff, documented in [installation.md](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#using-nix). Probably just got copied over to NixOS

jokesper:
> It's for the MacOS stuff, documented in [installation.md](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#using-nix). Probably just got copied over to NixOS

I didn't mean in our installation, I meant in the nixos flake upstream docs.
I couldn't find any mention of this option existing.

slotThe:
> > It's for the MacOS stuff, documented in [installation.md](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#using-nix). Probably just got copied over to NixOS
> 
> I didn't mean in our installation, I meant in the nixos flake upstream docs. I couldn't find any mention of this option existing.

Ah, I don't think I've ever seen any docs regarding that, no. It still seems to be [in nixcpp](https://github.com/NixOS/nix/blob/master/src/libfetchers/git.cc#L188) though

jokesper:
This is probably a case of forgotten documentation, which causes the feature to be less known, even by other contributors. I.e. it will receive less testing.
Though this is entirely speculation.

slotThe:
Perhaps you're right, yeah. Though it also is the only way to use flakes with git submodules at all (the support is *not great*) so I would imagine many people are at least peripherally aware of it

PR #966, opened 2025-02-07
Fixed exit on untranslateable keycode on macos

### LINKED ISSUE #817: [macos] Cannot translate from mac keycode: (12,568)
I have this error message from time to time since I installed my new `Unicorne` split keyboard (it's a Corne basically).
When it happens kmonad shuts down.
```
Encountered error in KeySource: [Cannot translate from mac keycode: (12,568)]
kmonad: [Cannot translate from mac keycode: (12,568)]
```
Any idea what could be wrong, could we just ignore the error instead of crashing the whole program?

jokesper:
We simply ignore keys we don't handle.
Maybe we should retransmit them instead?

This
- fixes #817

@jsimonetti, could you verify this works?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -23,7 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Fixed `tapMacro` and `tapMacroRelease` behaviour which was slightly broken in #873 (#906)
 - Fixed keycode translation problem on windows (#894)
 - Fixed keyrepeat not working in tty on linux (#913)
-- Fixed unrecognized keycodes on macos closing KMonad (#966)
+- Fixed unrecognized events on macos closing KMonad (#966)
 
 ## 0.4.3 – 2024-09-11
 
--- a/changelog.md
+++ b/changelog.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Fixed `tapMacro` and `tapMacroRelease` behaviour which was slightly broken in #873 (#906)
 - Fixed keycode translation problem on windows (#894)
 - Fixed keyrepeat not working in tty on linux (#913)
+- Fixed unrecognized keycodes on macos closing KMonad (#966)
 
 ## 0.4.3 – 2024-09-11
 
--- a/src/KMonad/Keyboard/IO/Mac/IOKitSource.hs
+++ b/src/KMonad/Keyboard/IO/Mac/IOKitSource.hs
@@ -80,12 +80,7 @@ iokitRead b = do
     peek $ b^.buffer
   case fromMacKeyEvent we of
     Nothing -> iokitRead b
-    Just (Left e)
-      | (`all` e) $ \case
-        NoMacKeycodeFrom _ -> True
-        _ -> False
-      -> do
-        for_ e (logDebug . displayShow)
-        iokitRead b
-      | otherwise -> throwIO e
+    Just (Left e) -> do
+      for_ e (logDebug . displayShow)
+      iokitRead b
     Just (Right k)  -> pure k
--- a/src/KMonad/Keyboard/IO/Mac/IOKitSource.hs
+++ b/src/KMonad/Keyboard/IO/Mac/IOKitSource.hs
@@ -80,4 +80,12 @@ iokitRead b = do
     peek $ b^.buffer
   case fromMacKeyEvent we of
     Nothing -> iokitRead b
-    Just e  -> either throwIO pure e
+    Just (Left e)
+      | (`all` e) $ \case
+        NoMacKeycodeFrom _ -> True
+        _ -> False
+      -> do
+        for_ e (logDebug . displayShow)
+        iokitRead b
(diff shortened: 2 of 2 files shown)


jsimonetti:
I will have a look. Thank you for a quick fix!

Please have some patience, my kmonad build is currently failing and I suspect something happend with nix's build support of submodules as my flake is no longer working as expected :(

jokesper:
Your problem may be related to #961. The solution for non macOs was dropping the submodule as it isn't needed. Though for macos you will need it.

jsimonetti:
Yes, that PR is indeed blocking me, somehow.

I managed to build it manually (with `nix build '.?submodules=1#default'` should anybody find this thread) and the result is 50% fixed...

kmonad no longer crashes when I click a mouse button, but it still crashes when I move it. The error messages is now:
```
==> /tmp/kmonad.stdout <==
connected
driver_connected 0
driver_version_mismatched 0
driver_connected 1
Encountered error in KeySource: [Cannot interpret mac switch: 2,Cannot translate from mac keycode: (1,49)]

==> /tmp/kmonad.stderr <==
kmonad: [Cannot interpret mac switch: 2,Cannot translate from mac keycode: (1,49)]
```

Weirdly enough, it doesn't crash when I scroll up, however, it does crash when I scroll down:
```
==> /tmp/kmonad.stdout <==
connected
driver_connected 0
driver_version_mismatched 0
driver_connected 1
Encountered error in KeySource: [Cannot interpret mac switch: 18446744073709551615,Cannot translate from mac keycode: (1,56)]

==> /tmp/kmonad.stderr <==
kmonad: [Cannot interpret mac switch: 18446744073709551615,Cannot translate from mac keycode: (1,56)]
```

jokesper:
Should be fixed now, though does the mouse still work?

jsimonetti:
> Should be fixed now, though does the mouse still work?

kmonad doesn't crash anymore, but the mouse stopped working ;)
weirdly enough, my trackpad still works, haha

jokesper:
Hmmm, fixing this would be a bigger change I think.
Since we should reemit any event we cannot process.
To do this we would probably need to "merge" reading and writing to devices, to allow a passthrough without making a KeyEvent out of it.
Alternatively we could alter the KeyEvent to an enum with an untranslateable case.

Maybe we should look at how to correctly catch only the keyboard and not any abitrary device.
I think `iokit-name` takes an optional product string, did you specify one?

jsimonetti:
No I did not. I have no idea how to get the actual iokit-name of my keyboard.
Also, do you know if i can bind it to multiple keyboards

jsimonetti:
image

Maybe this is also part of the issue. The mouse is recognised as a keyboard!?

jokesper:
> Also, do you know if i can bind it to multiple keyboards

The common way to do this is to start KMonad for each device. Under linux I use a udev rule for this.
I don't know the mac equivalent.

> Maybe this is also part of the issue. The mouse is recognised as a keyboard!?

Interesting. Which application is that. Our logic for this is in [`monitor_kb` of `keyio_mac.hpp`](https://github.com/kmonad/kmonad/blob/7784e590405429f5105d9d038cbe13f887160805/c_src/mac/keyio_mac.hpp#L150-L211) I think.

jsimonetti:
I will look into running KMonad for each device. For now, I don't think there's much you can do about a mouse being recognized as a keyboard!

That's the `Customise Modifier Keys` dialog box of `System Settings`.

jokesper:
Since this doesn't fix the issue, and just breaks things, I will close this PR

weavDarwin:
> Also, do you know if i can bind it to multiple keyboards

Sorry to chime in on an old closed thread, but I was poking around the links from #961 and @jsimonetti here's an example from my kmonad/nix-darwin config:

```
environment.etc = {
    ...
    "kmonad-internal.cfg" = {
        ...
        (defcfg
          input  (iokit-name "Apple Internal Keyboard / Trackpad")
          output (kext)
          fallthrough true
        )
        ...
    };

    "kmonad-hotdesk.cfg" = {
        ...
        (defcfg
          input  (iokit-name "Dell KB216 Wired Keyboard")
          output (kext)
          fallthrough true
        )
        ...
    };
  };
```

> I have no idea how to get the actual iokit-name of my keyboard.

Fetch input names with `list-keyboards` (the `kmonad.packages.aarch64-darwin.list-keyboards` flake output hasn't worked in at least a year for me but it's easy enough to build adhoc).

Then depending on whether I'm on my lap or at a desk with a shitty keyboard, I start kmonad with `sudo kmonad [/etc/kmonad-internal.cfg|kmonad-hotdesk.cfg]`. If you need to run it for multiple inputs simultaneously, idk how to do that.

jokesper:
> Fetch input names with list-keyboards (the kmonad.packages.aarch64-darwin.list-keyboards flake output hasn't worked in at least a year for me but it's easy enough to build adhoc).

Hmm, could you create an issue with the error.
Looking at the code, I think it may be related to the includes or maybe the compiler isn't called at all?

weavDarwin:
Opened on in #986.

PR #971, opened 2025-02-13
Add `tap-hold-next-press` variant

### LINKED ISSUE #951: Feature request: tap-hold-next-press
There are already a lot of tap/hold options, but the functionality of this one seems impossible to achieve from the existing ones... at least as far as I have been able to tell. 
- `tap-hold` comes close, but triggers hold functionality even if the next event after key press is release of a previously pressed key. This is annoying as I've got `space` mapped to a tap/hold button and when I am typing quickly I sometimes try inserting a space before I release the last key from the preceding word. 
 
`tap-hold-next-press` would do exactly as `tap-next-press`, only with a sort of timeout like `tap-hold`. For example:
```
(defsrc a)
(deflayer 
  some-layer
  @map) 
(defalias 
  map (tap-hold-next-press 200 a lsft))
```
Would produce the following functions:
```
      Ta Tb -> ab
Pa Pb Ra Rb -> B
   Pa Tb Ra -> B
Pb Pa Rb Ra -> b
      Tb Ta -> ba
```
and most importantly,
```
  Pa 199 Ra -> a
  Pa 201 Ra -> 
```

Since `tap-next-release` has got the 'hold equivalent' `tap-hold-next-release`, I think it seems reasonable that `tap-next-press` should get the same. 

And thank you to the developers for making such an brilliant and well-designed application :)

jokesper:
This fixes #951

### Description

The missing `tap-hold-next-press` variant.
@mistekko could you try this before I merge this?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Added `boot.initrd.services.kmonad.enable` NixOS option to use KMonad in the initrd (#941).
 - Added `key-seq-delay`, a more general version of `cmp-seq-delay`, which enforces a minimum delay
   after each key event. (#908)
+- Added `tap-hold-next-press` which is like `tap-next-press` but with an additional timeout. (#971)
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


mistekko:
Seems to work as expected. Thank you!

PR #974, opened 2025-02-18
Fix `multi-tap` cancelling on release of other keys

### LINKED ISSUE #973: With #958, rolling Space-Caps does not produce hold behaviour
Thank you very much for the implementation of #958, which resolves what I reported in #955! I compiled it today. I really appreciate your support here.

I use a layout with `mod4-or-ins (multi-tap 200 caps S-ins)` since today. I noticed another odd behaviour I wanted to report. Please excuse me if this is intentional behaviour I am unaware of.

On my layout (noted, a Neo2 variant) `caps` is a special modifier called `mod4`. Pressing e.g. `caps + l` produces `-`.

**The issue I want to report is** that pressing and holding `space`, then pressing and holding `caps`, then very quickly releasing `space` and then very quickly pressing `l` does not produce `-` but the letter assigned to `l`. In simpler words: Rolling from `space` to `caps` and then pressing `l` often fails to trigger the combination.

Any other key would work too: very quickly rolling from e.g. `f` to `caps`, followed by `l` fails to produce `-` but produces the letter.

This movement occurs very naturally during touch typing.

jokesper:
This fixes #973

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Fixed keycode translation problem on windows (#894)
 - Fixed keyrepeat not working in tty on linux (#913)
 - Fixed `multi-tap` not holding (#958)
+- Fixed `multi-tap` cancelling on release of other keys (#974)
 
 ## 0.4.3 – 2024-09-11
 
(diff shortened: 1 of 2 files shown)


slotThe:
@MaxG87 can you confirm that this fixes the issue for you?

MaxG87:
Yes, this fixes #973. Thank you very much!

Issue #820, opened 2024-02-21
Shifted symbols don't work in Rust applications


bermell:
Thanks for KMonad, it's a remarkable software! 

I've noticed that Rust GUI applications ([wezterm](https://github.com/wez/wezterm), [alacritty](https://github.com/alacritty/alacritty) and [neovide](https://github.com/neovide/neovide) so far) all have problems with shifted symbols in layers before the tab-hold-next-release timeout happens. 

[My config](https://github.com/bermell/bermell_linux_config/blob/main/kmonad/config.kbd) includes this: 

```
 spcsym (tap-hold-next-release 300 spc (layer-toggle symbols))
...
(deflayer symbols
  _      !     ..... 
...
)
```
i.e. 
tap spacebar -> get a space. 
hold spacebar, tab q -> get `!` 

It other non-rust programs this works as intended without issues. But in the rust programs, when I type fast I'll get a `1` when I wanted `!`. however, hold the spacebar, wait for a short while and then tap q, I get the intended `!`.  The behavior is the same for all symbols that you need to hold shift to get on a traditional keyboard. for example: 
* `2` where I wanted `@` 
* `3` where I wanted `#` 

etc... 

Feels like Rust is more to blame than kmonad but I want to raise it here to get some pointers on where to continue the investigation. 

I'm on the latest Ubuntu LTS (Ubuntu 22.04.4 LTS) with the latest release of kmonad (0.4.2). 

At my desk I use Corne keyboard with ZMK connected with bluetooth that works without issues in the Rust apps.

mdSlash:
I encountered the same issue when using macro-shifted keys in Kitty. I resolved it by adding [pause](https://github.com/kmonad/kmonad/blob/70a5e97518c87ff52be4b403d774e88c5c61e3c1/keymap/tutorial.kbd#L498C4-L498C13) before the shifted keys.

bermell:
> I encountered the same issue when using macro-shifted keys in Kitty. I resolved it by adding [pause](https://github.com/kmonad/kmonad/blob/70a5e97518c87ff52be4b403d774e88c5c61e3c1/keymap/tutorial.kbd#L498C4-L498C13) before the shifted keys.

Thanks for sharing your experience! I'm currently using Kitty and it doesn't have the issue describe though I also had to add some pauses to the macros I use. Based on what you said I tried defining a new button for `!` in the following ways by none them worked

```
(defalias
  exl (tap-macro P5 !)     ;;  just doesn't work
[...]

PR #975, opened 2025-02-19
Change default `key-seq-delay` to 1ms

jokesper:
This should fix most problems users may have, while still being not to
noticeable in `tap-macro`s. We may also want to consider changing from
milliseconds to microseconds or allowing fractional values for
`key-seq-delay` and `cmp-seq-delay`.

This effectivly partially reverts https://github.com/kmonad/kmonad/commit/c2ce56137a867b530bfe7fb3ccc6a0ea47b7a9c4:
> Set default for `key-seq-delay` to 5ms

(Sorry, my bad. I should have allowed for discussions via a PR.)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -21,6 +21,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 - `multi-tap` now holds when interrupted by another key while held. (#897)
 - `key-seq-delay` defaults to 1ms (#975).
+  If this slowdown in e.g. `tap-macro`s bothers you, you may want to set it to 0.
+  If you used `cmp-seq-delay`, you probably no longer have to,
+  as `key-seq-delay` effectivly already implies a `cmp-seq-delay`.
 
 ### Fixed
 
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 ### Changed
 
 - `multi-tap` now holds when interrupted by another key while held. (#897)
+- `key-seq-delay` defaults to 1ms (#975).
 
 ### Fixed
 
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -16,6 +16,7 @@
     - [Q: How to use the special features printed on Apple function keys?](#q-how-to-use-the-special-features-printed-on-apple-function-keys)
 - [General](#general)
     - [Q: Why can't I remap certain (non US) buttons?](#q-why-cant-i-remap-certain-non-us-buttons)
+    - [Q: Why don't shifted symbols work (sometimes)?](#q-why-dont-shifted-symbols-work-sometimes)
     - [Q: What does `Ta`, `Pa` and `Ra` stand for?](#q-what-does-ta-pa-and-ra-stand-for)
     - [Q: Why doesn't the 'Print' keycode work for my print screen button?](#q-why-doesnt-the-print-keycode-work-for-my-print-screen-button)
     - [Q: Why can't I remap the Fn key on my laptop?](#q-why-cant-i-remap-the-fn-key-on-my-laptop)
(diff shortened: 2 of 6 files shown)


jokesper:
> We may also want to consider changing from milliseconds to microseconds

I implemented this locally, and found with a `0.1` ms no difference in the outputted character speed in comparison to `1`ms. This may be due to the scheduler.
For three lorem-ipsums with a call to date before and after, I found about 3.3s with both `1`ms and `0.1`ms, about 14s for `5`ms and `0.1`s for no delay.

If you want to test around it yourself, you could try [my branch `frac-delays`](https://github.com/jokesper/kmonad/tree/frac-delays).

slotThe (commented):
> We may also want to consider changing from milliseconds to microseconds

I don't think this is necessary, the most important thing is that this actually still fixes the issue (in which case I think that 1ms is an acceptable compromise). @bermell could you perhaps test this PR?

jokesper:
for me it still fixes it in fcitx5 and my compositor, which both had problems.
Maybe we should have some kind of tracking issue, where we list other software which has said issue and the corresponding bug report. Or maybe a page in the currently empty wiki.

bermell:
> @bermell could you perhaps test this PR? 

Is there an artifact somewhere or should I look into compiling the code? :)

jokesper:
yes, in the CI pipeline.though the docker ci is only run on push. I. e. you can find it when you look at the branch and then the last ci pipeline run. und docker linux you can find a build artifact

jokesper:
Here is a direct link (though you need to be logged in): https://github.com/jokesper/kmonad/actions/runs/13411665712/artifacts/2615767659

bermell:
> Here is a direct link (though you need to be logged in): https://github.com/jokesper/kmonad/actions/runs/13411665712/artifacts/2615767659

with :point_up: ( `0.4.3, commit 671b7fc3ad9b23ccd2ae974cc6d0514477afa959`) everything worked as expected as long as I had `key-seq-delay 5` in my config. But if I removed that I got the behavior I described in #820 - the issue I created and recently closed.  Does that answer the question?

jokesper:
Could you test with lower values of this option? I.e. `2`, `3`, `4`, ....
This PR should still fix the desired behaviour while keeping the delay as low as possible.

bermell:
> Could you test with lower values of this option? I.e. 2, 3, 4, ....
This PR should still fix the desired behaviour while keeping the delay as low as possible.

Using `0.4.3, commit 671b7fc3ad9b23ccd2ae974cc6d0514477afa959` I can replicate the behavior with `key-seq-delay` = `1` and `2` but not with `3`. Does that help?

jokesper:
yes thank you.
It seems if we actually want to fix it by default the delay would always be quite noticeable with longer tap-macros. It's probably system and load dependent.
With such big differences (1ms vs 3ms) we may want to instead have some kind of wiki entry, where we list the problematic software with a possible upstream issue\*. The other alternative would be leaving it at 5ms giving people a slow done for their tap-macros and keeping working around the issues.
@slotThe what do you think?

\* a wiki entry would need to be moderated somehow, to find the real causes / enough to reproduce it. I don't have much experience with wikis on GitHub yet.
An alternative would be some kind of tracking issue.

slotThe:
> yes thank you. It seems if we actually want to fix it by default the delay would always be quite noticeable with longer tap-macros. It's probably system and load dependent. With such big differences (1ms vs 3ms) we may want to instead have some kind of wiki entry, where we list the problematic software with a possible upstream issue*. The other alternative would be leaving it at 5ms giving people a slow done for their tap-macros and keeping working around the issues. @slotThe what do you think?
>
> * a wiki entry would need to be moderated somehow, to find the real causes / enough to reproduce it. I don't have much experience with wikis on GitHub yet. An alternative would be some kind of tracking issue.

I think that having the default at either 1 or 0ms sounds best if we can't find a reasonable value that works for everyone. Instead of a wiki, perhaps a FAQ entry (linking to e.g. the original issue) would suffice. Since this can be caused by a variety of different factors, it doesn't sound like we should put a lot of effort into tracking this down in every individual case

jokesper:
> I think that having the default at either 1 or 0ms sounds best if we can't find a reasonable value that works for everyone. Instead of a wiki, perhaps a FAQ entry (linking to e.g. the original issue) would suffice. Since this can be caused by a variety of different factors, it doesn't sound like we should put a lot of effort into tracking this down in every individual case

I agree. I would take the 1ms amount to have it mostly fixed. Though 0 would also be perfectly fine with this much extra documentation we now have.
@slotThe if you don't have any preferred value for it, I would now merge this.

Issue #954, opened 2025-01-14
Stepped button keeps definition through layer-switch


jfmcbrayer:
For reference, this is from master at 25de8837fdd89b5c70bb313188f7430fe29e0c42, on Windows 11.

I have a configuration with three layers: QWERTY, Colemak, and a layer on held Windows/Meta key which really only has a stepped layer switch key:

``` clojure
(defalias toLS (tap-next-release met (layer-toggle layer-selector)))
(defalias ltog (stepped (layer-switch colemak) (layer-switch default)))
```

(full config and debug output will be below).

The intent is that Win-space will toggle between the QWERTY and Colemak layers. What actually happens is that on Win-space, the base layer correctly switches to Colemak. But what happens is that the next press of space switches the base layer back to default, even though Win is no longer pressed, and the layer-selector layer has been popped.

I am not sure whether this behavior of `stepped` is intentional or not. I can see an argument for either way, but it would be much more useful for me if I could have the stepped behavior only in the layer I bound it in. I could get access to both layers by binding two different keys in the layer-selector layer rather than trying to have one do both jobs, but having it on one key would be nice.

The full config:

``` clojure
(defcfg
    input (low-level-hook)
  output  (send-event-sink 500 30)
  fallthrough true
  allow-cmd false)

(defalias cbk (tap-next-press bspc lctl))
(defalias cen (tap-next-press enter rctl))
(defalias spa (tap-next-release spc lalt))
(defalias toLS (tap-next-release met (layer-toggle layer-selector)))
(defalias ltog (stepped (layer-switch colemak) (layer-switch default)))

(defsrc
    esc  `  1    2    3    4    5    6    7    8    9    0    -    =    bspc
    tab     q    w    e    r    t    y    u    i    o    p    [    ]    \
    caps    a    s    d    f    g    h    j    k    l    ;    '    enter
    lsft      z    x    c    v    b    n    m    ,    .    /    rsft
    lctl    lmet lalt           spc            ralt menu rctl
    )

(deflayer default
    esc  `  1    2    3    4    5    6    7    8    9    0    -    =    bspc
    tab     q    w    e    r    t    y    u    i    o    p    [    ]    \
    @cbk    a    s    d    f    g    h    j    k    l    ;    '    @cen
    lsft      z    x    c    v    b    n    m    ,    .    /    rsft
    lctl    @toLS lalt           @spa            ralt cmp rctl
    )

(deflayer layer-selector
[...]

PR #976, opened 2025-02-26
Fixed `multi-tap` holding while released and cancel the sequence

jokesper:
We also cancel `multi-tap` on release but only while released. This should fix `multi-tap` in a layer keeping it's definition through layer-switch (see #954 for an equivalent issue for `stepped`).

Also note, this does not revert #958 or #974.
Effectively they overshot the goal slightly.
It also doesn't reintroduce the issues #958 and #974 where designed to fix.

Hey @MaxG87, since you opened those issues originally and use `multi-tap` more heavily, could you test this PR and whether it matches expectations of what `multi-tap` should do.


MaxG87:
I'll do next week.

MaxG87:
I ran the changes for a complete afternoon and couldn't find any issues. Everything worked as expected.

Issue #494, opened 2022-03-02
Keyboard not recognized by kmonad (mac)


tmelz:
I have a Kinesis Advantage, when I run `make; ./list-keyboards` I get
```
Apple Internal Keyboard / Trackpad
Bluetooth USB Host Controller
(null)
```
The `null` only appears when the Kinesis is plugged in. Strangely, if I just use `input (iokit-name)` (without any string), it remaps my internal keyboard (expected), but the Kinesis is **not** remapped. 

Any ideas on how to resolve this?

MuhammedZakir:
Can you check whether Karabiner-Elements detect your device? Either in the devices tab or in the log tab. You don't need to update/change the dext for this.

tmelz:
Yes, it appears to detect it, log below. Under devices it says "No product name (No manufacturer name)"

```
[2022-03-02 10:13:26.026] [info] [observer] matching device:
[2022-03-02 10:13:26.026] [info] [observer]   registry_entry_id: 4294971006
[2022-03-02 10:13:26.026] [info] [observer]   vendor_id: 1523
[2022-03-02 10:13:26.026] [info] [observer]   product_id: 7
[2022-03-02 10:13:26.026] [info] [observer]   location_id: 0x14220000
[2022-03-02 10:13:26.027] [info] [observer]   is_keyboard: true
[2022-03-02 10:13:26.031] [info] [observer]   is_pointing_device: false
[2022-03-02 10:13:26.031] [info] [observer] rescan devices...
[2022-03-02 10:13:26.033] [info] [observer] (vendor_id:0x5f3, product_id:0x7) (device_id:4294971006) is observed.
[2022-03-02 10:13:27.033] [info] [grabber] caps lock is found on (vendor_id:0x5f3, product_id:0x7)
[2022-03-02 10:13:27.034] [info] [virtual_hid_device_service] virtual_hid_device_service_clients_manager (pointing) client is removed (size: 0)
[2022-03-02 10:13:27.035] [info] [grabber] (vendor_id:0x5f3, product_id:0x7) (device_id:4294971006) is grabbed.
[2022-03-02 10:13:27.035] [info] [grabber] caps_lock_led_state_manager is started: (vendor_id:0x5f3, product_id:0x7)
[2022-03-02 10:13:27.035] [info] [virtual_hid_device_service] virtual_hid_device_service_clients_manager (pointing) client is removed (size: 0)
```

MuhammedZakir:
[...]

PR #980, opened 2025-03-13
Set keyboard string for unknown keyboards on macos

jsimonetti:
### Description

When Macos cannot detect a keyboards' name (usually due to bad firmware), it sets the name to "Unknown External Keyboard". By mimicing this behavior in Kmonad, the keyboard can still be used. It will also allow usage of the input name in your config using `input (iokit-name "Unknown External Keyboard")`

### Checklist

### DIFF
--- a/c_src/mac/keyio_mac.hpp
+++ b/c_src/mac/keyio_mac.hpp
@@ -88,7 +88,7 @@ void open_matching_devices(char *product, io_iterator_t iter) {
         CFStringRef cfcurr = (CFStringRef)IORegistryEntryCreateCFProperty(curr, CFSTR(kIOHIDProductKey), kCFAllocatorDefault, kIOHIDOptionsTypeNone);
         if(cfcurr == NULL) {
             print_iokit_error("IORegistryEntryCreateCFProperty");
-            continue;
+            cfcurr = CFStringCreateWithCString (NULL, "Unknown External Keyboard", kCFStringEncodingUTF8);
         }
 
         // any device named "Karabiner ..." should be ignored
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Added `key-seq-delay`, a more general version of `cmp-seq-delay`, which enforces a minimum delay
   after each key event. (#908)
 - Added `tap-hold-next-press` which is like `tap-next-press` but with an additional timeout. (#971)
+- Set unknown keyboard names to "Unknown External Keyboard" on Macos. (#980)
 
 ### Changed
 


jokesper:
Since I don't have a Mac I sadly cannot test this.

Though from the docs I couldn't find the reason for when `IORegistryEntryCreateCFProperty` returns `NULL`. Is this only the case for keyboards (1) where the product string cannot be read (2)?

If for example only (2) is the case, this could lead to accidentally capturing non keyboard devices.

jokesper:
Closing this should fix #494.

jsimonetti:
> Since I don't have a Mac I sadly cannot test this.
> 
> Though from the docs I couldn't find the reason for when `IORegistryEntryCreateCFProperty` returns `NULL`. Is this only the case for keyboards (1) where the product string cannot be read (2)?
> 
> If for example only (2) is the case, this could lead to accidentally capturing non keyboard devices.

That is an excellent question, which I cannot answer with authority.

I CAN tell you that when I run the `c_src/mac/list-keyboards` program I get the following output:

```sh
jeroens-mac:mac root# ./list-keyboards
Apple Internal Keyboard / Trackpad
YubiKey OTP+FIDO+CCID
(null)
G502 HERO Gaming Mouse
USB Keyboard
CORSAIR K70 RGB MK.2 LOW PROFILE Mechanical Gaming Keyboard
Karabiner DriverKit VirtualHIDKeyboard 1.7.0
```

This corresponds to what I see in the `System Settings.app` under the `Customise Modifier Keys` dialog box, with the exception that the `(NULL)` entry is now called "Unknown External Keyboard.

image

It does not show the USB hub, microphone or webcam that are connected. Probably as they are not HID devices, though I am unsure and I realize this is anecdotal evidence.

Hopefully people from the linked issue can give this PR a try. It 'works for me' though :)

jokesper:
Hmm, I guess a working solution with maybe bugs, is better than a non working solution.

This would only change the behaviour if you don't specify a product string, which is already kind of buggy (see #817, ...).
I will wait for another person to test this before merging.

Since this is also the problem with #937 could also you test it?

jsimonetti:
AFAICT, when looking at https://github.com/kmonad/kmonad/blob/838f8ec2ad33e84094ea13f50af2d6e0a8137511/c_src/mac/keyio_mac.hpp#L163, there is an explicit match done for devices of type `kHIDUsage_GD_Keyboard`, which would prevent the case of capturing non-keyboard devices.

jsimonetti:
> Since this is also the problem with #937 could also you test it?

Just to clarify, would you like my to test if #937 solves my issue or if that PR works together with this PR?

jokesper:
> Just to clarify, would you like my to test if https://github.com/kmonad/kmonad/pull/937 solves my issue or if that PR works together with this PR?

Just testing it, as it is a PR which sits open due to it not being tested.
testing it together with yours is also fine. I don't think it would solve your issue, but if it does, it would also be great.

>  , there is an explicit match done for devices of type kHIDUsage_GD_Keyboard, which would prevent the case of capturing non-keyboard devices.

Oh right, great!

PR #982, opened 2025-03-18
Fixed permission error when opening symlinked device on Linux as non root

### LINKED ISSUE #981: Permissions when starting Kmonad with systemd
I am trying to map my caps lock key to the alt gr key, using a file called caps-to-altgr.kbd. When running `sudo kmonad $HOME/.config/kmonad/caps-to-altgr.kbd`, it works.

However, I cannot get it running with systemd despite trying everything the manual in https://github.com/david-janssen/kmonad/blob/master/doc/faq.md#making-a-launcher says.

This is my systemd service file:

`[Unit]
Description=KMonad keyboard remapping
After=graphical.target

[Service]
ExecStart=/usr/bin/kmonad /home/jonas/.config/kmonad/caps-to-altgr.kbd
Restart=always
RestartSec=3
Environment=DISPLAY=:0
Environment=XDG_SESSION_TYPE=wayland
Environment=XDG_RUNTIME_DIR=/run/user/%U

[Install]
WantedBy=default.target
`

I also made a rule in `/etc/udev/rules.d`: `KERNEL=="uinput", MODE="0660", GROUP="uinput", OPTIONS+="static_node=uinput"`

`ls -l /dev/input/` says it belongs to groups root and inout. however, `ls -l /dev/input/by-id/my-keyboard-kbd´ actually gives  `lrwxrwxrwx 1 root root`. Could that be the issue?

Anyway, when trying to debug the error with `journalctl --user -u kmonad -f`, i get `kmonad: /dev/input/my-keyboard-kbd: openFd: permission denied (Permission denied)

What am I missing?

jokesper:
Fixes #981

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 - Fixed keyrepeat not working in tty on linux (#913)
 - Fixed `multi-tap` not holding (#958, #976)
 - Fixed `multi-tap` cancelling on release of other keys (#974)
+- Fixed permission error when opening symlinked device on Linux as non root (#982)
 
 ## 0.4.3 – 2024-09-11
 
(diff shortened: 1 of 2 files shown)


jokesper:
This doesn't fix the issue (symlink lookup is already done implicitly, we don't need to do it)

PR #992, opened 2025-04-20
Placeholder and aliases in `defsrc`

### LINKED ISSUE #990: Feature request: a dummy symbol or prefix for defsrc
Some keyboards/remotes/combos have multiple keys mapped to the same keycode. Currently `defsrc` requires each symbol to be unique, i.e. no symbol can appear multiple times. Therefore we can only define one of the keys and ignore the other.

For example, if I have a keyboard that has this physical layout (a real world example at the end):
```
a s d f f d 3 s a del
```
My `defsrc` can only be:

```
(defsrc
 a s d f 3 del
)
```
which is visually too different from the physical layout.

Can we have some dummy prefix so we can do something like this, to maintain the layout:
```
(defsrc
 a s _d _f f d 3 _s _a del
)
```
and maybe a pre-processor that strip anything with leading `_` before moving on? 

Or simply
```
(defsrc
 a s _ _ f d 3 _ _ del
)
```
assuming we implement something to ignore this special duplication?

The same goes for `deflayer`, for symmetry.

This can also help ignore those special keys mentioned in
https://github.com/kmonad/kmonad/issues/487

A real world example: I have a wireless remote control that has two sides of keypads; the back side is a full keyboard (well almost full); the front side has some selected keys (arrow keys, enter, etc.) from the full version.

Thanks for this great project!

jokesper:
Added placeholder and aliases in `defsrc`. Furthermore I reworked `joinAliases`, this leads to order irrelevance in `defalias`.

Resolves #990

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -289,9 +289,8 @@
   these buttons, to keep the actual `deflayer` statements orderly.
 
   A `defalias` can contain any number of aliases, and it can refer backwards or
-  forwards to layers without issue. The only sequencing that needs to be kept in
-  mind is that a `defalias` cannot refer forward to another `defalias` that is
-  not yet defined.
+  forwards to layers or other aliases without issue. Thoug you may not create
+  cyclic aliases.
 
   Here we define a few aliases, but we will define more later. Notice that we
   try to only use 3 letter names for aliases. If that is not enough to be clear,
@@ -303,6 +302,8 @@
 (defalias
   num  (layer-toggle numbers) ;; Bind num to a button that switches to a layer
   kil  C-A-del                ;; Bind kil to a button that Ctrl-Alt-deletes
+  foo @bar                    ;; An alias with a forward reference to another alias
+  bar a
 )
 
 
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -250,9 +250,18 @@
   quite nicely (although wider columns will allow for more informative aliases,
   see below).
 
+  You can also insert a placeholder button using `XX`. This is almost the same
+  as extra white space but you still need to bind a button to it. Said button is
+  never used. This is occasionally useful if you have buttons which don't emit a
+  keycode or have the same keycode multiple times on the same keyboard. In such
+  cases large chunks of whitespace may be more confusing then a placeholder
+  button. If you further want to specify which button this placeholder refers
+  to, you can use an alias (see below).
+
   Instead of keycodes directly you can also use an alias (see `defalias`).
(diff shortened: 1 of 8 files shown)


jokesper:
The `show` on the internal keycode representation is not ideal, but it's easier than adding source annotations or reconstructing the source. Not having it would decrease the quality of the error message.

keymap/tutorial.kbd
```diff
@@ -279,6 +292,8 @@
   kp0    kp.
 )
 
+(defsrc :name with-aliases @foo @power)
```
tsankuanglee, line 295: Could we add a newline so it's more consistent with other examples? 
```
(defsrc :name with-aliases
 @foo @power
)
```

keymap/tutorial.kbd
```diff
@@ -303,6 +317,9 @@
 (defalias
   num  (layer-toggle numbers) ;; Bind num to a button that switches to a layer
   kil  C-A-del                ;; Bind kil to a button that Ctrl-Alt-deletes
+  power XX                    ;; Placeholder for power button of keyboard
+  foo @bar
+  bar a
```
tsankuanglee, line 322: Can we add a comment to explain the forward reference?

tsankuanglee (approved):
Tested this feature branch:

1. alias forward referencing works
2. alias in `defalias` works

Small documentation suggestion but non-blocking.

Thank you for the wonderful work!

keymap/tutorial.kbd
jokesper: This was more or less intentional. The reason was it being a `defsrc` with only one button.
This aids as an example of small `defsrc`s and also as an example that the initial newline is actually not required.
It's mostly a reminder that white space is irrelevant.
Furthermore it's a rather irrelevant `defsrc`, it's not used, has only one button and the one button which it has is picked abitrarly.

That's why I would personally prefer the one-line variant there.

All that aside, this exists only as a test case and demonstration for aliases and placeholders in `defsrc` (the prior arguments where not about existence but styling). I couldn't think of how to build this into the other `defsrc`. If you have a more "common" example I would be happy to replace it.

keymap/tutorial.kbd
jokesper: yes, makes sense.

keymap/tutorial.kbd
jokesper: @tsankuanglee If you don't have any alternative example or some other suggestion, I would like resolve this thread and merge it

tsankuanglee: Ah, sorry about that -- please go ahead and merge.

Issue #567, opened 2022-06-13
Apply layout to all available keyboards if you omit the input device


oblitzitate:
It would be convenient for `kmonad` to scan all available keyboards in `/dev/input` if you don't specify an input device. 

I use a laptop with an external keyboard, but sometimes like to unplug and carry it around. Such a feature would make it convenient so that I don't have to rely on a script to iterate through all keyboard devices.

I believe `xremap` does this pretty well when you omit the `--device` option. It lets you know which keyboards failed to grab and which are selected.

david-janssen:
Have you checked the `/dev/input/by-id` subdirectory? It usually contains a symlink to your keyboards by an unchanging id, so no scanning required.

oblitzitate:
Yes, but since I use multiple keyboards (e.g. laptop keyboard, usb keyboard 1, usb keyboard 2, etc.), I still have to run `kmonad` for each keyboard. Note: laptop keyboards are located in `/dev/input/by-path`.

Obviously, I can easily do this all with a script, but it'd be nice if it was built-in where I'd just call `kmonad` once and don't need to put an input device in `defcfg`.

oblitzitate:
I don't know if this is the ideal way to do it, but this is the script I use:

```bash
#!/usr/bin/env bash
kbd_file="$1"
fd '^(usb|platform).+event-kbd$' /dev/input | while read keyboard; do
  kmonad -w 0 --input "device-file \"$keyboard\"" "$kbd_file" &
done
```

maueroats:
Suggestion: put a link to this information in the tutorial by the `device-file` Linux information. 
Use the link to [obliterate's comment](https://github.com/kmonad/kmonad/issues/567#issuecomment-1154233223). 

The command-line arguments were not very discoverable to me. I think changing the help string for `--input` would help a bit. It you could add "Specifies the `device-file`" to make it a little clearer. Or call the value `DEVICE` or even `INPUT_DEVICE` instead of `ITOKEN`.

Related:

You could consider a section in the FAQ: "What command line arguments are available for kmonad?"

david-janssen:
> Suggestion: put a link to this information in the tutorial by the device-file Linux information.

Pullrequests are welcomed :)

tobiasBora:
If anyone has a script to make it reload when a new keyboard is plugged, please let me know ;-)

tsankuanglee:
[...]

PR #993, opened 2025-04-21
add input/output parameter format to faq

tsankuanglee:
Add `--input`/`--output` format requirement to FAQ.

credit goes to https://github.com/kmonad/kmonad/issues/567#issuecomment-1154233223

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -23,6 +23,7 @@
     - [Q: Why do some key combination not work?](#q-why-do-some-key-combination-not-work)
     - [Q: When I run KMonad I get error `Not available under this OS`](#q-when-i-run-kmonad-i-get-error-not-available-under-this-os)
     - [Q: Where can I find a list of keycodes which can be used in KMonad?](#q-where-can-i-find-a-list-of-keycodes-which-can-be-used-in-kmonad)
+    - [Q: How do I provide the input and output from the command line?](#q-how-do-i-provide-the-input-and-output-from-the-command-line)
 
 <!-- markdown-toc end -->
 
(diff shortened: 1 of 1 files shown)


jokesper:
Thanks! I would change a few things before merging this (mostly formatting), though since they effect quite a bit (of the change) and PRs on PRs are weird on GitHub I will paste my changes here. If you don't have anything against them, I will apply them and merge it.

Without syntax highlighting, since the backticks did funny things:
    diff --git a/doc/faq.md b/doc/faq.md
    index 5e0c64d..9f2b677 100644
    --- a/doc/faq.md
    +++ b/doc/faq.md
    @@ -296,17 +296,26 @@ A: List of keycodes can be found [here.](https://github.com/kmonad/kmonad/blob/m
     
     ### Q: How do I provide the input and output from the command line?
     
    -A: The input/output can be assigned via the `--input`/`--output` (or the short style `-i`/`-o`) options to override the value in the `.kbd` config file. These options take exactly the same assignment as the input/output in the `defcfg` block. For example, if we have this line in our `.kbd` config file:
    +A: The input / output can be assigned via the `--input` / `--output`
    +(or the short style `-i` / `-o`) options to override the value in the `.kbd`
    +config file. These options take exactly the same assignment as the `input` /
    +`output` statement in the `defcfg` block. For example, if we have this line
    +in our `.kbd` config file:
     
    -```
    +```clojure
       input  (device-file "/dev/input/by-id/my-kbd")
     ```
    -Then the way to call kmonad is:
    -```
    -bash$ kmonad --input 'device-file "/dev/input/by-id/my-kbd"' my-kbd.kbd
    -```
    -Notice that single quotes are used to pack the `device-file` and the path into one string, and also note that the path needs to be wrapped with double quotes. If your shell does not support single quotes, you can also escape the double quotes.
    +We may instead call KMonad using:
     
    +```bash
    +kmonad --input 'device-file "/dev/input/by-id/my-kbd"' my-kbd.kbd
     ```
    -$ kmonad --input "device-file \"/dev/input/by-id/my-kbd\"" my-kbd.kbd
    +
    +Notice that single quotes are used to pack the `device-file` and the path into
    +one string, and also note that the path needs to be wrapped with double quotes.
    +If your shell does not support single quotes, you can also escape the double
    +quotes.
    +
    +```bash
    +kmonad --input "device-file \"/dev/input/by-id/my-kbd\"" my-kbd.kbd
     ```

tsankuanglee:
Great edits! Feel free to edit however you see fit. Thanks a lot!

jokesper:
Thanks!

PR #996, opened 2025-05-07
tutorial: Note that a broken kmonad does not break other keyboards

notevenaperson:
One word change.

On all suported platforms, Kmonad processes [initiate bound](https://github.com/kmonad/kmonad/blob/master/doc/quick-reference.md#input-and-output) to capture input from a specific keyboard, so when the process gets to an unusable state only the process' keyboard is affected (except on Windows).

### Description

Include a description for your changes, including the motivation behind them.

### Checklist

### DIFF
--- a/keymap/tutorial.kbd
+++ b/keymap/tutorial.kbd
@@ -734,7 +734,7 @@
   switch into a layer that you can never get out of. Or worse, you could
   theoretically have a layer full of only `XX`s and switch into that, rendering
   your keyboard unusable until you somehow manage to kill KMonad (without using
-  your keyboard).
+  this keyboard).
 
   However, when handled well, `layer-switch` is very useful, letting you switch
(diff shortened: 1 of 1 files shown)


jokesper:
> except on Windows

this also affects `(iokit-name)` (leaving the name empty captures every keyboard). Furthermore we probably want to support multiple devices in the future, so clarifying it's about the specify inputs not just a single (but also not every) is probably the correct solution.

jokesper:
how about "said" instead of "this" or "your"?

notevenaperson:
Oh I didn't think singular keyboard support wasn't that clear cut. 

Since you still want to merge a change like this, I've used "said" as: 

> ...rendering a keyboard unusable until you somehow manage to kill KMonad (without using said keyboard).

I've just made a separate pull request with this change.

jokesper:
Closed as fixed in #997.

Thanks!

PR #997, opened 2025-05-08
tutorial: Note you could kill kmonad with other keyboards

notevenaperson:
### Description

A small wording tweak.

Usually kmonad has been configured by the user to be bound to capture input from a specific keyboard, so when the process gets to an unusable state only the process' keyboard is affected.

Based on #996.
### Checklist


jokesper:
Hmmm, the `a keyboard` feels a bit of, I will replace it with `the configured keyboard` and merge it. I don't really want to argue about such small details so I'm cutting the discussion here short.

Thanks!

Issue #952, opened 2025-01-08
Cannot translate from windows keycode: 223


KyrLance:
I'm a long time kmonad user on linux, for work I now need a windows laptop, so I compiled kmonad for windows, as per installation instructions.

Using my keymap, everything works except the $ dollar sign, which gives me: "Encountered error in KeySource: Cannot translate from windows keycode: 223".

I'm using swiss german layout on windows itself.

I'm really out of ideas right now. The error seems to be from the defsrc itself
Probably someone knows a solution here?

Cheers

jokesper:
KMonad internally uses linux keycodes, for mac or windows, we translate them via a huge table. It seems, that for VK_OEM_8 we couldn't find out what key it were.
I assume you use a config you used before on linux, which key did you use there?
If you are interested, the line is: [317 in `Windows/Types.hs`](https://github.com/kmonad/kmonad/blob/master/src/KMonad/Keyboard/IO/Windows/Types.hs#L317)

KyrLance:
Hi there, thanks for your reply! Yes, I'm using my linux config: https://github.com/kmonad/kmonad-contrib/blob/master/keymaps/Kyrill/kymap.kbd
And just changed the input and output config as described in the docs.

The key that handles $ in swiss layout is the backslash \ key on the defsrc. Pressing that one will give me the error. On swiss layouts, \ is typed with right alt + <, but aliasing that also won't work. I guess it's literally the key itself that is the issue on windows?

jokesper:
Hmm, we already translate `VK_OEM_5` to `\`.
According to the docs:
> it can vary by keyboard

Though some keys like `VK_OEM_PLUS` has:
> For any country/region, the + key

This leads me to believe that we capture keys after localization.
This is inconsistent with the rest of KMonad. We should probably look for a way to capture them before remapping.

Furthermore we capture keys in userland which causes issues with other applications that do the same.
In general the windows keyio bit is not very good.

jokesper:
You may want to take a look at [Kanata](https://github.com/jtroo/kanata) it started as a reimplentation of KMonad and has diverged since. Based on it's documentation I think it's far more windows focused.
This may be a valid solution while the KMonad windows support is only working partially

KyrillGobber:
I did see kanata, but did not realize it was a reimplementation of kmonad and wanted to stick to kmonad. I really like kmonad a lot.

I understand that windows support is no priority, though, so I'm gonna try kanata for now.

Thanks a lot for your replies!

[...]

PR #1005, opened 2025-06-07
Handle empty reads on windows

### LINKED ISSUE #1003: Encountered error in KeySource: Cannot translate from windows keycode: 6357060
I installed KMonad on Windows 11 24H2 using the release 0.4.4 binary. I am using the Polish (Programmers) Windows keyboard layout. The keyboard is an SPC Gear GK550 Omnis Kailh Red RGB.

My `config.kbd` file looks like this:

```
(defcfg
  ;; For Windows
  input  (low-level-hook)
  output (send-event-sink 500 30)

  ;; This option tells KMonad to let non-configured keys act normal
  fallthrough true
)



(defsrc
    a    s    d    f    g    h    j    k    l    ;
)


(defalias
    alt_a (tap-hold-next-release 800 a lalt)
    met_s (tap-hold-next-release 800 s lmet)
    ctl_d (tap-hold-next-release 800 d lctl)
    sft_f (tap-hold-next-release 800 f lsft)

    sft_j (tap-hold-next-release 800 j rsft)
    ctl_k (tap-hold-next-release 800 k rctl)
    met_l (tap-hold-next-release 800 l rmet)
    alt_; (tap-hold-next-release 800 ; ralt)
)



(deflayer homerowmods
    @alt_a   @met_s   @ctl_d   @sft_f   g   h   @sft_j   @ctl_k   @met_l   @alt_;
)
```

To replicate the error, I can run KMonad by opening a Terminal inside of the directory containing `kmonad.exe` and `config.kbd` and executing
```
.\kmonad.exe .\config.kbd
```

The error does not always occur: I would say there are periods where it does not occur at all, and then periods when I can't run KMonad without the error appearing.

The issue appears to be related to #952 

Is there any fix other than trying out kanata?

jokesper:
Fixes #1003 (probably)

### DIFF
--- a/c_src/keyio_win.c
+++ b/c_src/keyio_win.c
@@ -84,12 +84,12 @@ LRESULT CALLBACK keyHandler(int nCode, WPARAM wParam, LPARAM lParam)
 }
 
 // Read an event from the pipe and write it to the provided pointer
-void wait_key(struct KeyEvent* e)
+DWORD wait_key(struct KeyEvent* e)
 {
   DWORD dwRead;
   ReadFile(readPipe, e, sizeof(e), &dwRead, NULL);
   //printf("receiving: %d\n", e->keycode);
-  return;
+  return dwRead;
 }
 
 // Insert the keyboard hook and start the monitoring process
--- a/changelog.md
+++ b/changelog.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 - Fixed excessive backtracking in the parser leading to unreadable errors (#985)
 - Fixed the `c_src/mac/list-keyboards.c` on Apple SDK < 12.0 (#987)
+- Fixed random crash on startup on windows. (#1005)
 
 ## 0.4.4 – 2025-04-11
 
--- a/src/KMonad/Keyboard/IO/Windows/LowLevelHookSource.hs
+++ b/src/KMonad/Keyboard/IO/Windows/LowLevelHookSource.hs
@@ -36,8 +36,24 @@ foreign import ccall "release_kb"
 -- | Pass a pointer to a buffer to wait_key, when it returns the buffer can be
 -- read for the next key event.
 foreign import ccall "wait_key"
-  wait_key :: Ptr WinKeyEvent -> IO ()
+  c_wait_key :: Ptr WinKeyEvent -> IO Word32
 
+-- | A small wrapper around the C function 'c_wait_key'. Sometimes the read from
+-- the internal pipe in the C code gets interrupted. Probably due to events we
+-- want to ignore. Therefore we limit retries on empty reads. (See #1003)
+wait_key :: HasLogFunc e => Ptr WinKeyEvent -> RIO e ()
+wait_key = wait_key' 3
+ where
+  wait_key' :: HasLogFunc e => Int -> Ptr WinKeyEvent -> RIO e ()
+  wait_key' 0 _ = throwIO TooManyEmptyReads
+  wait_key' n buffer = do
+    read <- liftIO $ c_wait_key buffer
+    if
+      | read == 0 -> do
+        logDebug "Empty read from key-source."
+        wait_key' (n-1) buffer
+      | fromEnum read == sizeOf (undefined :: WinKeyEvent) -> pure ()
+      | otherwise -> throwIO $ UnexpetedNumberOfBytesRead read
 
 --------------------------------------------------------------------------------
 
@@ -78,7 +94,6 @@ llClose ll = do
 -- NOTE: This can throw an error if the event fails to convert.
 llRead :: HasLogFunc e => LLHook -> RIO e KeyEvent
 llRead ll = do
-  we <- liftIO $ do
-    wait_key $ ll^.buffer
-    peek $ ll^.buffer
+  wait_key $ ll^.buffer
+  we <- liftIO $ peek (ll^.buffer)
   either throwIO pure $ fromWinKeyEvent we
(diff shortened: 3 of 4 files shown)


notevenaperson:
I won't be able to say if the patch really makes a difference right now. There's randomness to this event and now I can't even get it to happen on a build of `master`.

notevenaperson:
When the event occurs during launch, having applied the patch:
```
Opening KeySink
Initializing Windows key sink
Opening KeySource
Registering low-level Windows keyboard hook
Launching process: emitter_proc
read bytes 0, receiving 0
Empty read from key-source.
```
And kmonad keeps working.

Unfortunately I have gotten this other message sometimes too, which is still a crash: 
```
Encountered` error in KeySource: Too many empty reads. If you get this error message internal assumptions are incorrect.
```

I'm still trying to get this last error to occur when I have the debug helpers enabled to be able to see what keycodes kmonad is receiving. 
This isn't quite enough.

jokesper:
The last error was intentionally added by me. I assumed the read would be interrupted by some event and since I thought it was a rare event, which was safe to ignore I added this check. What could cause the read to be interrupt three times in a row?

jokesper:
Ok, I'm now more confused than before the printed error message says: "No error"?! what? I'm not sure the patch is applied correctly (so for sending diffs btw). I added a commit with the patch. Could you retest the single event failure?

notevenaperson:
The way I presented the logs can be confusing. The log with "No error" is the patch from the *first* diff you sent here. 

Though to reassure us both I've pulled the commit and will post a new log

notevenaperson:
Is it normal that the "`[ OK ] 8 bytes read:`" line appears when I start kmonad in the Powershell host (a terminal) but doesn't appear when I redirect kmonad's debug output to a file?

Your log:
```
Opening KeySink
Initializing Windows key sink
Opening KeySource
Registering low-level Windows keyboard hook
Launching process: emitter_proc
Empty read from key-source.

Received event: Press <d>
Running hooks
Registering untimed hook: 1
Emitting: Press <d>

```

The reason the error doesn't cause kmonad.exe to exit is definitely something in my environment, it's an odd enviroment. I'm using the Powershell script below. It calls kmonad in a loop and appends the instances of all the logs to a file, which I can search later for error messages. I only learned a bit Powershell literally yesterday. It has like all the features of `sh`, and they _look_ similar, but there's a bunch of  hidden gotchas because of the NT way of doing things.

```pwsh
param ($KmonadBin)
$i=0

do {
	$i = $i + 1
	Write-Host "Starting n=$i"
	# We use ArgumentList to pass the variable to the script block
	$job = Start-Job -ScriptBlock { & $args[0] -o "send-event-sink" "-i" "low-level-hook" "$($Env:Userprofile)\.config\kmonad\yea.kbd" "--log-level" "debug" 2>&1 >> "C:\Users\XXX\desktop\pullnunupatch.txt" } -ArgumentList $KmonadBin
	Start-Sleep 4
	Write-Host "Slept 4"
	Stop-Job -Job $job
} while ($true)
```

jokesper:
> Is it normal that the "`[ OK ] 8 bytes read:`" line appears when I start kmonad in the Powershell host (a terminal) but doesn't appear when I redirect kmonad's debug output to a file?

This probably has something todo with `printf` in C vs in Haskell.
sadly this and the the later error message are both C `printf` which do the debugging I needed.
I was interested in what the return of `ReadFile` was in the case it read 0 bytes.

I will forward the data to the haskell side and print it properly from there. I will write again when I'm done.

Side note:
The high CPU usage is why I limited the retries on empty reads. If they don't go away after a few tries we may be in an infinite loop actively waiting. Since the abitrary limit of three also caused issues, I was wondering what exactly is causing this (i.e. read the error message).

notevenaperson:
`
Encountered error in KeySource: Could not read from low-level hook. Read aborted with error: 0x6
`

!!

notevenaperson:
False alarm, it does not happen *every time*. 'thwas just bad luck

jokesper:
This should work now. Please give it a try again.

notevenaperson:
Something I notice right away is the debug log no longer logs any key press events. I'm not sure that's intended. 

There's also a noticeable lag between pressing "A" on the keyboard and seeing "a" on screen now while kmonad is running.

On the other hand, I got no empty read-related error messages after having launched kmonad 100 times, using the powershell script.

jokesper:
could you test the parent commit? i.e. 181cee0

notevenaperson:
> could you test the parent commit? i.e. [181cee0](https://github.com/kmonad/kmonad/commit/181cee0f48981c486099656115c0dd279f11f767)

We're back. Now testing 🫡

notevenaperson:
I have no logged errors in the 163 separate times I've launched kmonad with the script that was previously used to produce errors.

jokesper:
Great. Then I will drop the buggy version and merge this in about half a day. If you notice any problems please keep me informed.

jokesper:
Do you still get the debug message `Empty read from key-source.`? If not, I would like to drop this edge case and also throw an error in that case. We shouldn't add workarounds for problems which are not observed.

jokesper:
> Something that I think would benefit from some kind of workaround right now are the instances where kmonad receives a legitimate keycode it cannot translate like https://github.com/kmonad/kmonad/issues/1003#issuecomment-2952658494 and https://github.com/kmonad/kmonad/issues/952.

The best solution would be to add all those keycodes (sadly this is hard in those cases).
I think the second best solution would be an option like `fallthrough` but for unknown key events. i.e.: `remit-unknown-events`. With some hacky code for `255` on windows, since we cannot reemit it but can instead not capture it.

I think this may be quite doable and will hack up a draft soon.

jokesper:
> I think this may be quite doable and will hack up a draft soon.

I'm not quite sure about the config options yet.

KMonad could differentiate between known keycodes, unknown keycodes and unknown events (e.g. mouse move).
Furthermore we may want to error, ignore, ignore but triggering hooks (i.e. `tap-next`), reemit or reemit and trigger hooks.
So we need to decide whether to error or not, whether to reemit or not and whether to block them in the queue.
For mouse movement there may be a further sensible option of triggering hooks but not being able to the delay them.

I'm not quite sure what kind of options to introduce allowing a clean design with the other options in the future.
Ideas are welcome.

PR #1008, opened 2025-07-01
Drop `Generic` from `Keycode` to reduce compile time

jokesper:
On my system this reduces the clean compile time (without dependencies) from 66.3s to 34.9s.

The extra `hashable` dependency was already an indirect dependency via `rio`.


jokesper:
Maybe having the keycode as a newtype around `Int` is the better choice, but since this would be a larger refactor, I will merge this solution instead.

PR #1022, opened 2025-09-23
Udev needs groups to be system groups to allow adding them

### LINKED ISSUE #1020: systemd 258 release breaks uinput group udev rules
The latest release of [systemd](https://github.com/systemd/systemd/releases/tag/v258) will not apply udev rules for groups that are not created as --system. This breaks the recommended way to acquire uinput  permissions in the FAQ.

I worked around it by creating a new group as follows:

`sudo groupadd --system uinput_sys`

and modifying the udev rule to use this group.

Then add your user to this group and logout to effect changes.

There may be a better way, but I am not sure.

jokesper:
Fixes #1020.

Thanks @mashtonian.

@slotThe should this be mentioned in the changelog? It's not a change in KMonad but in the install instructions.
Though it may still result in breakage for users updating systemd.

Furthermore I'm unsure how to fix the nixos module. Should we hardcode a gid?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0)
 
 ## Unreleased
 
+### Breaking
+
+- With systemd v258 udevd ignores `GROUP=` setting with a non-system group.
+  If you followed the [FAQ entry: How do I get Uinput permissions?](doc/faq.md#q-how-do-i-get-uinput-permissions)
+  and created the `uinput` group delete the group (`sudo groupdel uinput`)
+  and create a new group with the user added to it (see FAQ entry).
+
+
 ### Added
 
 - `XX` may be used in `defsrc` as a placeholder button.
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -40,7 +40,7 @@ If the `uinput` group does not exist, create a new group with:
 1. Make sure the `uinput` group exists
 
 ``` shell
-sudo groupadd uinput
+sudo groupadd --system uinput
 ```
 
 2. Add your user to the `input` and the `uinput` group:


slotThe (commented):
I think a changelog entry might be good, if only to make people aware that they need to adjust their groups. 

As for the nix module, I suppose we just need to set a `gid` [thereabouts](https://github.com/kmonad/kmonad/blob/master/nix/nixos-module.nix#L183), right? Perhaps remove `hardware.uinput.enable` and replicate the tiny [uinput.nix](https://github.com/NixOS/nixpkgs/blob/ef6712a91b2cd6b2e30551c8b92e79efd543a31f/nixos/modules/hardware/uinput.nix) along the way (which we sort of do already anyways)

jokesper:
After looking through the nixpkgs source code I found [the part where the gids are allocated](https://github.com/NixOS/nixpkgs/blob/ca40fe77e621d237f5aeaa1c2accf635c4a2ad39/nixos/modules/config/update-users-groups.pl#L79)
starting at 999 and decreasing towards 400. So they are already system groups.
Duplicating `hardware.uinput.enable` (`nixos/modules/hardware/uinput.nix`) doesn't seem like a good idea, since we would then have two uinput groups doing the same thing.

> which we sort of do already anyways

Yes, but for the initrd bit. since there is no `boot.initrd.hardware.uinput.enable` but this part is unaffected anyways, since it didn't make use of a uinput group (due to the need for a hardcoded nonconflicting gid).

I think we should leave the nixos bit as is and optionally remove the line `users.groups.uinput = { };`.

slotThe:
Oh, indeed `getent group uinput` returns `uinput:x:990:slot`, so it actually seems like the nixos module is fine!

PR #1023, opened 2025-09-28
Add definitions for Japanese-specific keys

shkentee:
### Description
Added definitions for previously undefined Japanese keys into  src/KMonad/Keyboard/IO/Windows/Types.hs
This update improves compatibility when using Japanese keyboard layouts.

### Checklist
- [x] I've read [CONTRIBUTING.md](https://github.com/kmonad/kmonad/blob/master/CONTRIBUTING.md)
- [ ] I've also appended my changes to the `CHANGELOG.md` if applicable (not applicable for keymap definitions only)

### DIFF
--- a/src/KMonad/Keyboard/IO/Windows/Types.hs
+++ b/src/KMonad/Keyboard/IO/Windows/Types.hs
@@ -175,11 +175,11 @@ winCodeKeyCodeMapping =
   , (0x15, KeyKatakana)
   , (0x15, KeyKatakanaHiragana)
   , (0x15, KeyHangeul)
-  -- , (0x16, ???)            -- Defined as VK_IME_ON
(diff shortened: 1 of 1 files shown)


jokesper:
I investigated this a bit in the context of #952. The best I could find out is that it's very hard to properly do the remapping we do on linux or mac and it's keyboard specific what oem_N is (see docs).
Since this is the second time something related to oem_N keys is requested, they should probably be added to the keycodes if only as windows only entry (similar to the mac ones).

`KeyHangeul` and `KeyHanja` are already defined but with different keycodes.

And in the case of `KeyKatakanaHiragana` which you want to add twice we may want to switch the selection option in the mapping from `maximum` / `minimum` to `first` / `last` so we can have an abitrary mapping. (Maybe a good change in general)

Issue #864, opened 2024-07-21
Unrecognized compose-char "stepped"


caph1993:
I am trying to use the `stepped` functionality explained in the [tutorial](https://github.com/kmonad/kmonad/blob/master/keymap/tutorial.kbd#L504) and the [quick reference](https://github.com/kmonad/kmonad/blob/master/doc/quick-reference.md#general-purpose-buttons). I copied and pasted, but kmonad reports a syntax error, as if `stepped` did not exist.

```
kmonad: Parse error at 208:11:
    |
208 |   ctl-lck (stepped (press-only lctl) (release-only lctl))
    |            ^
Unrecognized compose-char
```

Is this a bug? Is there a workaround? Thanks in advance.

I'm using the latest binary, with version `0.4.2, commit 3da3165c413225d924a55933b723ff8ee00f9231` obtained with the command `./kmonad -V`.

caph1993:
I found the problem. The `stepped` functionality is too recent, and is not present in the latest released version (0.4.2).

I built the latest (unreleased) version following the [docker-based build instructions](https://github.com/kmonad/kmonad/blob/master/doc/installation.md#using-docker), and it worked!

It was as simple as  running these three commands:

```sh
docker build -t kmonad-builder github.com/kmonad/kmonad.git
docker run --rm -it -v ${PWD}:/host/ kmonad-builder bash -c 'cp -vp /root/.local/bin/kmonad /host/'
docker rmi kmonad-builder 
```

jokesper:
Hmmm, this mismatch between docs and code is somewhat unfortunate, a solution would having a release and master branch.
(Spekulation)
Though this was probably tried, evidient by the fact that a develop branch exists, but people (including me) simply make their pr against the default branch or something.

Maybe a link to the docs at a specific commit?

caph1993:
Or maybe writting in the docs "supported since version x.x.x" for those that are already supported, and something like "(only supported in dev version)". And updating those comments during each release.

slotThe:
I think it's quite natural that the docs on `master` should be for the development version. For matching documentation there are always the respective release tags

jokesper:
> I think it's quite natural that the docs on `master` should be for the development version.

True, but the problem lies in navigating to the docs.
Simply using the link to the tutorial on [kmonad/kmonad](github.com/kmonad/kmonad) would result on docs not matching the release.
Originally I was thinking something like this (and maybe this is the correct solution):
```diff
[...]

PR #1035, opened 2025-12-07
Add annotations since when features exist in tutorial.kbd

### LINKED ISSUE #1033: ignore-missing directive triggers kmonad parse error
Hi!

I discovered today the :ignore-missing  directive from the quick-reference
I decided to add it but if I do, kmonad fails to parse

(defcfg
   input (device-file "/dev/input/by-id/usb-Logitech_USB_Keyboard-event-kbd" :ignore-missing true)
   output (uinput-sink "Logitech k120")
)

```
kmonad -d my_keyboard.kbd


kmonad: Parse error at 2:78:
  |
2 |    input (device-file "/dev/input/by-id/usb-Logitech_USB_Keyboard-event-kbd" :ignore-missing true)
  |                                                                              ^
unexpected ':'
expecting ')'
```

I don't get what I'm doing wrong, could you help me please?

Linux 24.04
kmonad 0.4.4, commit 09c304851b1088d80661599ee3daceb617837be9

jokesper:
Since I remember you not wanting to put a full link in the README we may want to instead include this information in the tutorial.
The quick-references are still missing. Sadly this doesn't really fit into the tutorial format maybe we want to change it.

Alternatively we may want to make a new release soon. Though this would still take some time before downstream catches up.
(If we do release now, we should move the karabiner change to the correct section).

(This was mostly written before looking up the ArchLinux package version and finding out they upgraded. Apparently they were just 3 Months behind us.)

This should reduce confusion on new features (or with outdated packages):
- fixes #1033

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,4 +47,4 @@ request:
 Please also remember to update the `CHANGELOG.md` file, as well as any other
 documentation that your pull request touches upon.  For example, when
 implementing a new feature, you should update the `tutorial.kbd` file
-accordingly.
+accordingly. Please also add a `(unreleased)` after the documentation.


slotThe:
> I found the relevant comment. Also I slightly misremembered / misrepresented the reason. For reference your reply to previous suggestion for adding a link to the docs of the newest release: [#864 (comment)](https://github.com/kmonad/kmonad/issues/864#issuecomment-2242632776)

Ah, I see now. I think my main objection was that if I saw something like

    [the configuration tutorial]: https://github.com/kmonad/kmonad/blob/0.4.2/keymap/tutorial.kbd

in a README I would not expect it to point to some tag. I guess getting users to use tags (and git) is probably too much to ask, though :) I think something like a link to the latest release documentation is fine, as long as it clearly states that. E.g.,

    [the configuration tutorial (HEAD)](keymap/tutorial.kbd)
    [the configuration tutorial (v0.4.2)]: https://github.com/kmonad/kmonad/blob/0.4.2/keymap/tutorial.kbd

or something of that nature

> Since we also want those tags for new features I thought about adding it to contributing.

:+1:

>
> It lists:
>
> > * Pull requests for new features should be made against the `develop` branch.
>
> I can't remember ever having something merged into develop. We should probably either remove that part or actually have this as our workflow. Though I think a separate develop branch would just complicate the workflow.

Oh yeah, `develop` goes way back and was abandoned not too long after it was born, AFAIR. We should definitely remove that.

jokesper:
I think the `(since x.y.z)` part is mostly not needed. It also clutters the tutorial somewhat. I think just having the `(unreleased)` sections would be enough.

slotThe:
No opinions from me, both approaches sound fine. Whatever you prefer

jokesper:
I think most of those annotation don't really interrupt. The only one I feel slightly weird about is the `(unreleased)` mid sentence on the aliases. But that wouldn't get fixed by dropping the `since` tags.

PR #1036, opened 2025-12-16
Create Apple_A1243_108_keys.kbd

aiamuzz:
Created an additional file specifically for the Apple 108 keys A1243 Model.

### Description

I have modified the 'iso_100.kbd' file to accomodate the addtional keys of this 108 Key Apple Keyboard.

### Checklist

- [x] I've read [CONTRIBUTING.md](https://github.com/kmonad/kmonad/blob/master/CONTRIBUTING.md)
- [ ] I've also appended my changes to the `CHANGELOG.md` if applicable

### DIFF
--- a/src/KMonad/Keyboard/Keycode.hs
+++ b/src/KMonad/Keyboard/Keycode.hs
@@ -881,6 +881,7 @@ keyNames = mconcat
 aliases :: Q.MultiMap Keycode Text
 aliases = Q.mkMultiMap
   [ (KeyEnter,            ["ret", "return", "ent"])
+  , (KeyKpEqual,          ["kp="])
   , (KeyMinus,            ["min", "-"])
   , (KeyEqual,            ["eql", "="])
   , (KeySleep,            ["zzz"])
@@ -924,7 +925,6 @@ aliases = Q.mkMultiMap
   , (KeySlash,            ["/"])
   , (KeyClear,            ["clr"])
   , (KeyNumLock,          ["nlck"])
-  , (KeyKpEqual,          ["kp="])
   , (KeyKpSlash,          ["kp/"])
   , (KeyKpEnter,          ["kprt"])
   , (KeyKpPlus,           ["kp+"])


aiamuzz:
Sorry, for the half baked cookies,

please pause this PR until i am able to get the other things in place, If I fail then somebody else can help setup this Apple 108 Keys keyboard beyond the kbd keymaps.

A learning curve, I am trying to get this to work on my fork to learn things.

src/KMonad/Keyboard/Keycode.hs
```diff
@@ -976,7 +984,7 @@ aliases = Q.mkMultiMap
   , (KeyNextSong,         ["nexttrack"])
   , (KeyPreviousSong,     ["previoustrack"])
   -- , (KeyStopCd,           ["stop"]) -- conflict with KeyStop
-  , (KeyEjectCd,          ["eject"])
+  , (KeyEjectCd,          ["eject", "ej"])
```
jokesper: Please add this as a new entry further above. This is in the part of hid names.
So we have two parts. They should be in the first section (sometimes there are comments which aren't commented out lines starting them)

aiamuzz:
> Oh, I see you already found it.

yeah, learning, am i on the right track ?

I did add the following Keys ...

KeyFnF13 to KeyFnF19 (which the Apple A1243 Model with 108 keys additionally has on it) and all Func keys are invoked with the Fn pressed ... hence I've added KeyFnFxx names beyond the standard 12.

do these additions f13 to f19 need any defining(function code) ?
If so It could be well above my paygrade @jokesper ...

keymap/template/Apple_A1243_108_keys.kbd
```diff
@@ -0,0 +1,35 @@
...
+
+  -------------------------------------------------------------------------- |#
+
+(defsrc
+  esc  f1   f2   f3   f4   f5   f6   f7   f8   f9   f10  f11  f12  ej    	f13  f14  f15	 	    f16  f17  f18  f19 
+  grv  1    2    3    4    5    6    7    8    9    0    -    =    bspc  	fn   home pgup  		clr  kp=  kp/  kp*
+  tab   q    w    e    r    t    y    u    i    o    p    [    ]    \     del  end  pgdn  		kp7  kp8  kp9  kp-
+  caps 	  a    s    d    f    g    h    j    k    l    ;    '    	ret                     		kp4  kp5  kp6  kp+
+  lsft       z    x    c    v    b    n    m    ,    .    /    	  rsft       	  up         	  kp1  kp2  kp3  ret
+  lctl lalt lmet           	    spc                 	  rmet ralt rctl    left down rght  		kp0  kp.
```
jokesper: Can Please align the keys better.
I don't think the kp keys are staggered like that. Though I didn't look it up.
Who has `fn` where the `ins` key should be?

jokesper (changes requested):

src/KMonad/Keyboard/Keycode.hs
```diff
@@ -524,6 +524,13 @@ data Keycode
   | KeyFnF10
   | KeyFnF11
   | KeyFnF12
+  | KeyFnF13
+  | KeyFnF14
+  | KeyFnF15
+  | KeyFnF16
+  | KeyFnF17
+  | KeyFnF18
+  | KeyFnF19
```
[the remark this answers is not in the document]
aiamuzz: > We already have keyf13 to keyf24. keyfnfXX are not hid keys and are also not in input-event-codes.h

oh , ok ... do you mean the keyf13 to keyf24 are sufficient even on KB's that have Fn involved in invoking them ?

Shall I delete those entries I've made or let them be, I just did that as the format keyfnfXX was found in the file.

keymap/template/Apple_A1243_108_keys.kbd
jokesper: This contains tab characters. This breaks the alignment (on GitHub) but also for different tab choices.

src/KMonad/Keyboard/Keycode.hs
jokesper: `kp=` is not the problem. The ejectcd alias as eject is here because those are the same key named differently by different standards. You are currently in the `HID Usage Names` section of the aliases (Header https://github.com/kmonad/kmonad/blob/master/src/KMonad/Keyboard/Keycode.hs#L959).

Adding a new entry similary to clear or kpequal further above in the alias table would be better, since this separates the reasons for why the aliases exist.
(For example directly below the other new aliases)

src/KMonad/Keyboard/Keycode.hs
```diff
@@ -910,7 +922,9 @@ aliases = Q.mkMultiMap
   , (KeyComma,            ["comm", ","])
   , (KeyDot,              ["."])
   , (KeySlash,            ["/"])
+  , (KeyClear,            ["clear", "clr"])
```
jokesper:
```suggestion
  , (KeyClear,            ["clr"])
```

KeyClear as clear alias is already handled by some dynamic approach above (`keyNames`, the drop 3 bit)

keymap/template/Apple_A1243_108_keys.kbd
aiamuzz: @jokesper ... 

As one can see the keys seem out of alignment, If I can correct the same using only spaces(without using tabs) it can be made visually appealing currently it is a bit of an eye sore

keymap/template/Apple_A1243_108_keys.kbd
aiamuzz: > This contains tab characters. This breaks the alignment (on GitHub) but also for different tab choices.

apologies for the messy PR conversation board, learning as i go the nuances of using github on my 1st real world PR, 

I have removed the 'tabs' and used sonly spaces keeping them consistent across the deflayer as well, but the layout is visually off, 

I can improve it visually by using more spaces wherever needed, shall I(of course i will keep it consistent across the deflayer as well ?

keymap/template/Apple_A1243_108_keys.kbd
```diff
@@ -0,0 +1,35 @@
...
+  used in place of `102d'. `lsgt', which stands for "less (<) / greater (>)",
+  is yet another valid name for that key.
+
+(deflayer name
+  _  _   _   _   _   _   _   _   _   _   _  _  _  _       _  _  _        _  _  _  _ 
+  _  _    _    _    _    _    _    _    _    _    _    _    _    _     _   _ _       _  _  _  _
+  _    _    _    _    _    _    _    _    _    _    _    _    _    _      _  _  _       _  _  _  _
+  _     _    _    _    _    _    _    _    _    _    _    _     _                           _  _  _  _
+  _       _    _    _    _    _    _    _    _    _    _        _           _             _  _  _  _
+  _ _ _                 _                    _ _ _      _ _ _       _  _
```
jokesper: Now they are even weirder aligned. Maybe something like this:

```suggestion
   _   _    _    _    _    _    _    _    _    _    _    _    _     _        _    _    _          _    _    _    _   
   _   _    _    _    _    _    _    _    _    _    _    _    _     _        _    _    _          _    _    _    _  
   _     _    _    _    _    _    _    _    _    _    _    _    _   _        _    _    _          _    _    _    _  
   _       _    _    _    _    _    _    _    _    _    _    _      _                             _    _    _    _  
   _         _    _    _    _    _    _    _    _    _    _         _             _               _    _    _    _  
   _   _    _                    _                      _    _      _        _    _    _          _    _
```

aiamuzz:
> Interesting. What happens with the f13 - f24 keys you originally added as fnf13... I think thei may be regular f keys. This is the main point why we can't fixup the PR. Currently this breaks every key after. Without we have no idea which keys should go there. On Mac it is probably does nothing since the translation table wasn't modified

Interesting, ok let me do some testing ... will report back ...

PR #1047, opened 2026-02-06
Minor corrections under docker install method

salman-farooq-sh:
Update docs for docker installation method.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -266,28 +266,29 @@ for your OS, etc. You can even use the provided `Dockerfile` for development
 testing. As of now, the built image is not meant to *run* `kmonad`, just to
 build it.
 
-Just do this from the `Dockerfile` directory:
+Just do this from the repository root directory after cloning:
 ``` shell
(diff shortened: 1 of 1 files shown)


jokesper (commented):
- Why introduce proper numbering while still referring to "the first one" instead of "step 1." or something.
- Does the url need the https:// prefix? it wasn't there before and makes the comment longer.

Though those are just some nitpicks.

salman-farooq-sh:
> Why introduce proper numbering while still referring to "the first one" instead of "step 1." or something.

Oh, I didn't think about it like that. It looks okay to me right now though. Let me know if you want me to create another commit.

> Does the url need the https:// prefix? it wasn't there before and makes the comment longer.

Yes, it wasn't working without the `https://` prefix on my docker installation which I am sure is a pretty recent one.

Thanks.

salman-farooq-sh:
Hey, sorry for the super late response but I just force-pushed a single combined commit for everything. Thanks.

jokesper:
Thanks!

(no problem)

PR #1053, opened 2026-03-29
WIP: Implement Coaround fixing Copilot key not mappable

### LINKED ISSUE #931: Bind copilot key to something else.
Abomination of windows. Copilot key is literally super+shift. I would be very grateful to get rid of this nonsense and have logical button option

jokesper:
A ruth sketch of an implementation. It should mostly work.

Problems:
1. Effectively tries to duplicate Sluice, Dispatch, Hooks in one component (use existing ones)
2. trying to press sft / met + copilot simultaniously causes problems (if it results in `Psft Pmet Pf23`, `Psft Pmet Pcopilot` instead of `Psft Pcopilot` is handled)
3. Always on. Even for users not having those keys (should have some opt-in `defkeys`? part in the config)
4. Requires lots of warnings on edge cases + documentation in general.
5. `copilot` without config option should be an error / warning.
6. `copilot` should not be allowed in `deflayer` (?)

Eventhough this would already fix #931 but 1, 3 and 4 would need some addressing first.
@AlejandroMinaya could you test whether it is handled properly. As said before some modifiers may cause issues.
Please list such cases

### DIFF
--- a/kmonad.cabal
+++ b/kmonad.cabal
@@ -54,6 +54,7 @@ library
     , unliftio
     , template-haskell
   default-extensions:
+      BangPatterns
       ConstraintKinds
       DeriveFunctor
       DeriveGeneric
@@ -70,6 +71,7 @@ library
       NamedFieldPuns
       NoImplicitPrelude
       OverloadedStrings
+      PatternSynonyms
       RankNTypes
       TemplateHaskell
       TupleSections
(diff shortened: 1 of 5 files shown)


slotThe (commented):
I'll be honest: I think this is a cool workaround, but I'm not sure that it's a good idea to take on all of this complexity, just to get what's essentially one rogue key to behave in a sane way…

jokesper:
Will close this since #1055 is merged

PR #1055, opened 2026-04-08
Alternative "solution" to #931

### LINKED ISSUE #931: Bind copilot key to something else.

jokesper:
The implemented "fix" in #1053 is actually relatively easy with the existing code base.
Simply using `tap-hold-next`. This adds a rather lengthly explanition of multiple workarounds depending on the need.
(and a starting off point in case somebody wants to do this properly outside of KMonad)

[Rendered](https://github.com/jokesper/kmonad/blob/doc-fix-931/doc/hardware-macro-keys.md)

This effectively fixes #931 though I would like @AlejandroMinaya testing of it

Also since this is a rather lengthy document (for an faq entry) proof reading should probably also be done.

Some core benefits of this approach:
- No extra code in KMonad to maintain
- Allows remapping copilot key only on some layers without the global delay it incorrs
- Explicit to the user that this is very hacky
- Easily extendable (no `defkeys` / hardcoding required)

Alternatives.
Instead of a lengthy detailed walkthrough I initially wrote a somewhat shorter FAQ entry:

---

The copilot key emits `Plmet Plsft Pf23` when pressed and `Rf23 Rlsft Rlmet` when released.
If any of the keys is already pressed / released while the copilot key is held
we don't get the corresponding modifier events.

If you want to map the Copilot key but don't care about the modifiers
(e.g. because tapping them does nothing special) you can simply map `f23`.

If you don't want the modifier events maybe because your config looks like this

```clojure
(defsrc lsft lmet)
(deflayer base
  #(d o n ' t)
  #(p r e s s spc m e)
)
```

you can use `(tap-hold-next 1 XX XX :timeout ...)` bound to those modifiers.
This causes press events followed by an event within 1 ms to be ignored.
(`(tap-hold-next-press 1 ... XX :timeout ...)` is arguably more correct).

The config with the Copilot key also mapped would then look like the following:

```clojure
(defsrc lsft lmet f23)
(deflayer base
  (tap-hold-next 1 XX XX :timeout #(d o n ' t))
  (tap-hold-next 1 XX XX :timeout #(p r e s s spc m e))
  #(c o p i l o t)
)
```

For other macro keys a similar approach can be used.

---

Though there are lots of other corner cases which should probably maybe be highlighted + I don't want to let the information of the key repeat oddity rod in some github issue comment

Furthermore I included the new document and the FAQ in the cabal dist which is used on e.g. ArchLinux to unpack to `/usr/share/doc/kmonad`. Though curiously on my (broken and outdated) dualboot this directory is empty. Will need to investigate this on a working system.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -16,6 +16,7 @@
     - [Q: How to use the special features printed on Apple function keys?](#q-how-to-use-the-special-features-printed-on-apple-function-keys)
 - [General](#general)
     - [Q: Why can't I remap certain (non US) buttons?](#q-why-cant-i-remap-certain-non-us-buttons)
+    - [Q: How can I remap the Copilot key / other hardware macro keys?](#q-how-can-i-remap-the-copilot-key--other-hardware-macro-keys)
     - [Q: Why don't shifted symbols work (sometimes)?](#q-why-dont-shifted-symbols-work-sometimes)
     - [Q: What does `Ta`, `Pa` and `Ra` stand for?](#q-what-does-ta-pa-and-ra-stand-for)
     - [Q: Why doesn't the 'Print' keycode work for my print screen button?](#q-why-doesnt-the-print-keycode-work-for-my-print-screen-button)
@@ -227,6 +228,13 @@ you would specify the letter that would be there in qwerty (`[`).
 This is of course also true for buttons which are in a different place.
 On the german layout the `+` is reached without shift, while on US qwerty it's
 on `S-=`. This means it's also not a valid keycode to specify in your `defsrc`.
+
+### Q: How can I remap the Copilot key / other hardware macro keys?
+
+By mapping the corresponding primary key (e.g. `f23`) and optionally
+dropping modifiers with `(tap-hold-next 1 XX XX :timeout ...)`.
+
+For a detailed explanation, see [Hardware Macro Keys](hardware-macro-keys.md).
 
 ### Q: Why don't shifted symbols work (sometimes)?
 
--- a/doc/hardware-macro-keys.md
+++ b/doc/hardware-macro-keys.md
@@ -1,0 +1,151 @@
+# Hardware Macro Keys
+
+Some keyboards have hardware macro keys.
+They are especially common with laptops shipping with some advertised Microsoft features.
+
+Some of those keys include:
+
+| Key                    | Macro                                  | Further References |
+|------------------------|----------------------------------------|--------------------|
+| [Copilot Key][copilot] | `Plmet Plsft Pf23`                     | #931               |
+| [Office Key][office]   | `Plmet Plalt Plctl Plsft` (unverified) |                    |
+
+[copilot]: https://www.microsoft.com/en-us/windows/learning-center/unlock-productivity-with-the-copilot-key
+[office]: https://support.microsoft.com/en-us/office/using-the-office-key-df8665d3-761b-4a16-84b8-2cfb830e6aff
+
+## Remapping
+
+There are multiple methods of remapping such hardware macro keys:
+
+- [Mapping a primary key](#mapping-a-primary-key).
+  The other key events are kept, and we need a "primary" (e.g. F23 for Copilot).
+- [Filtering out non-primary modifiers](#filtering-out-non-primary-modifiers).
+  This fixes the issue of the other key events being kept but still needs a "primary".
+- [Multi-layer solution](#multi-layer-solution).
+  Slightly improves accuracy. Wanted if we don't have a "primary" key (e.g. with the Office key).
+- [Filtered multi-layer solution](#filtered-multi-layer-solution).
+  Same improvements as [Filtering out non-primary modifiers](#filtering-out-non-primary-modifiers)
+  but without the need for a "primary" key.
+
+There are further improvements to the detection possible
+which would increase the complexity in KMonad too much
+or are not worth implementing. If you want to try yourself
+against this problem, there are some improvement ideas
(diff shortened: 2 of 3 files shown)


AlejandroMinaya:
I confirm that this minimal example works for me. I only had to change `:timeout` for `:timeout-button`. I will adapt my configuration and report back.

AlejandroMinaya:
I'm tweaking my config but I notice that when I press the `lmet` while holding `f23`, only the release event is printed in the kmonad log. Only when I press it again does it behave as intended. Is it possible that kmonad is not differentiating between `lmet` vs `met` when the keypress comes from the copilot key?

Here's the log
```
Opening KeySink
Registering Uinput device
Opening KeySource
Initiating ioctl grab
Launching process: emitter_proc

Received event: Press <met>
Running hooks
Registering untimed hook: 2
Registering 1ms hook: 1

Received event: Press <sft>
Running hooks
Registering untimed hook: 3
Registering untimed hook: 4
Block level set to: 1
Registering untimed hook: 6
Registering 1ms hook: 5

Received event: Press <f23>
Running hooks
Registering untimed hook: 7
Registering untimed hook: 8
Registering untimed hook: 9
Unblocking input stream, no stored events
Injecting event: Press <f23>
Registering untimed hook: 10

Rerunning event: Press <f23>
Tried cancelling expired hook: 1
Running hooks
Registering untimed hook: 11
Registering untimed hook: 12
Registering untimed hook: 13
Registering untimed hook: 14
Block level set to: 1
Registering untimed hook: 16
Registering 200ms hook: 15
Tried cancelling expired hook: 5
Cancelling hook: 15
Pushed layer to stack: numbers
 1. numbers
Base-layer: qwerty

Registering untimed hook: 17
Unblocking input stream, no stored events

Received event: Release <met>
Running hooks
Registering untimed hook: 18
Registering untimed hook: 19
Registering untimed hook: 20
Registering untimed hook: 21

Received event: Press <met>
Running hooks
Registering untimed hook: 22
Registering untimed hook: 23
Registering untimed hook: 24
Registering untimed hook: 25
Block level set to: 1
Registering untimed hook: 27
Registering 1ms hook: 26
Cancelling hook: 26
Registering untimed hook: 28
Unblocking input stream, no stored events
Emitting: Press <0>
0
Received event: Release <met>
Running hooks
Registering untimed hook: 29
Registering untimed hook: 30
Registering untimed hook: 31
Registering untimed hook: 32
Emitting: Release <0>

Received event: Release <f23>
Running hooks
Popped layer from stack: numbers
Base-layer: qwerty

Registering untimed hook: 33
Registering untimed hook: 34

Received event: Release <sft>
Running hooks
```

jokesper:
`met` is an alias for `lmet` sadyl copilot key presses the left meta key. We don't get any press event for lmet because it is already pressed by copilot. There is a slight difference in the behaviour possibly allowing us to differentiate between them: we don't get repeat events for the lmet pressed by copilot but do if pressed manually. Sadly this would require more logic which would need to be maintained. The other fix shouldn't handle this better (to my knowledge). This is also outlined at the end of the proposed hardware macro keys.

AlejandroMinaya:
I understand. Since the use case doesn't justify the needed complexity, this seems like a natural stop point for now. Thank you for the thorough look into this.

slotThe:
I pushed some typo fixes as a fixup commit, as this is perhaps easier to digest than doing this in a review

EDIT: Oops, I picked the wrong commit as the base :) But this can be fixed when merging and squashing manually

jokesper:
Thanks a bunch. Maybe I should better (or at least once after it is done) proof read notes I write at 2 am.
Will do so tomorrow.

Side note about the alignment of the markdown table:
Alignment makes it easier to read and comprehend when viewing the unrendered version but also causes the potential issue
of greatly increasing change areas when new entries might be added making it harder to review.

The result is we end up with something like the following (taken from `Joiner.hs`):
```haskell
-- | Joining Config
data JCfg = JCfg
  { _cmpKey  :: Button  -- ^ How to prefix compose-sequences
  , _implArnd :: ImplArnd -- ^ How to handle implicit `around`s
  , _kes     :: [KExpr] -- ^ The source expresions we operate on
  }
makeLenses ''JCfg
```

Formatting commits are also quite unpleasant when it comes to merging multiple PRs in the same region (e.g. the others need to solve merge conflicts caused by formatting changes). Though I think this is manageable.

Benefit is of course how easy it is to see, that there is a table column missing for the office key (Further References is missing closing `|`)

In this case (especially for in reference to the existing formatting of KMonad) aligning is probably the better choice.

AlejandroMinaya (approved):

slotThe:
I agree that this can be a problem in code sometimes, though with `git diff --word-diff=color` (or delta or magit) I think it's still worth it to align most things -- humans are pretty good at pattern recognition, and alignment seems to help at least this human.

For docs in particular, I don't think reviewing huge diffs will be much of a burden, and I personally have a pretty hard time reading things like

``` markdown
| Key | Macro | Further References |
|-----|-------|--------------------|
| [Copilot Key][copilot] | `Plmet Plsft Pf23` | #931 |
| [Office Key][office] | `Plmet Plalt Plctl Plsft` (unverified) |
```

as plain text in my editor (which is how I read the majority of markdown files; call me oldschool, I guess :))

jokesper:
It seems Archlinux creates a /usr/share/doc/kmonad but it is empty. Perhaps something we could consider in the future but I will adjust the commit message body.
(I will also fixup the commits) I think it's ready to merge then.

PR #1056, opened 2026-04-15
Fix copy paste error in error message

jokesper:
Played around with trying to break kmonad using minal config. Turns out there was a copy paste error (probably) in the error reporting.


slotThe:
thanks!

